PR #5, opened 2024-08-24
Added support for Docker

YannickFricke:


mgerhold:
What about the backend? Is it possible to have both run in separate docker containers? Or should the backend also be added here?

YannickFricke:
Currently the Docker container only includes the frontend - so the backend would need another Dockerfile.

Usually you create independent containers so you can scale them on their own :)

mgerhold:
> Currently the Docker container only includes the frontend - so the backend would need another Dockerfile.
> 
> Usually you create independent containers so you can scale them on their own :)

I don't think that the aspect of scaling is important here, since those containers should be used for developing. To the main focus should be on ease of use.
For deployment, we'll have to see how we do it later on.

My favorite solution would be a single container that contains all the necessary tools to run both applications. But the repositories should be mounted from outside so that they reside on the user's system. I don't have the time (nor the knowledge) to look into this.

PurpurTentakel97:
I think it was verry nice you did than but we would need that as a development environment.
So the solution that @mgerhold is suggesting seems good to me.
I will close this PR now.

When this branch is changing feel free to reopen in.

Issue #12, opened 2024-08-27
footer text


PurpurTentakel97:
add a description text in the footer. 
we should discuss if we hand to hardcode this text or load it from the database.
when we load it from the database every year could have its own description.

PR #13, opened 2024-08-30
9 resposive header and footer

### LINKED ISSUE #9: resposive header and footer
need to make the footer and header responsive for mobile.

PurpurTentakel97:
close #9 
see #12

### DIFF
--- a/src/elements/header.svelte
+++ b/src/elements/header.svelte
@@ -81,7 +81,7 @@
 		justify-content: space-between;
 		padding-top: 0.5rem;
 		padding-bottom: 0.5rem;
-		border-bottom: solid 3px black;
+		border-bottom: solid 1px black;
 	}
 
 	ul {
--- a/src/elements/header.svelte
+++ b/src/elements/header.svelte
@@ -38,9 +38,9 @@
 			<LogoSmall classList="header-mobile-img" />
 
 			<button class="hamburger" on:click={toggleOpen}>
-				<div></div>
-				<div></div>
-				<div></div>
+				<div class={isOpen ? 'rotated' : 'normal'}></div>
+				<div class={isOpen ? 'rotated' : 'normal'}></div>
+				<div class={isOpen ? 'rotated' : 'normal'}></div>
 			</button>
 		</div>
 		{#if isOpen}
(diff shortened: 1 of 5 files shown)


src/elements/header.svelte
```diff
@@ -1,34 +1,81 @@
...
+	export let menu;
+
+	let isOpen = false;
+
+	function toggleOpen() {
+		isOpen = !isOpen;
+	}
+
+	onMount(() => {
+		const colapse = () => {
```
mgerhold, line 15: typo
```suggestion
		const collapse = () => {
```

src/elements/header.svelte
```diff
@@ -1,34 +1,81 @@
...
+			isOpen = false;
+		};
+
+		window.addEventListener('scroll', colapse);
```
mgerhold, line 19: I think this is a cool idea, but when interacting with the website it felt a bit weird to me. I would not expect the menu to close (maybe because I'm not used to it, judging from other websites). Maybe we should ask someone with more experience with UI/UX?

mgerhold (changes requested):
My impressions and findings so far:

- In the mobile view, the black line under the header is missing. That can lead to the header and footer "becoming one".
image
- In general I'd suggest making that black border smaller (maybe only 1 or 2 pixels).
- I'd like a smaller offset when jumping to an anchor mark. I know that you don't want the heading to be "smashed" against the top, but at the moment I think the offset is a bit too big.
- When using a mobile phone in landscape mode, the hamburger menu doesn't fit the screen. I think we either need a scrollbar or two columns.

src/elements/header.svelte
PurpurTentakel97: My idea was that when you scroll your focus is clearly on the side behind the menu.
So collapsing the menu was logical to me.  maybe even when you choose a menu point. what do you think?

src/elements/header.svelte
PurpurTentakel97: done

src/elements/header.svelte
PurpurTentakel97: But I agree. On Desktop it feels verry weird.
Maybe one would need to try it on mobile at this point.

mgerhold: I talked to the chat about this. We all think that closing the menu on scroll is okay, but only if the scrolling starts outside of the expanded menu. But when the cursor (or finger, on mobile) starts dragging on the menu, the menu should not be closed.

Another idea is: If you start scrolling on the expanded menu and you move the cursor/finger upwards, then the menu could collapse accordingly. And if you keep dragging until less than half of the menu is still visible, then the menu could "auto-collapse" the rest of the animation.

peddermaster2: I have to agree with @mgerhold

mgerhold (changes requested):
The hamburger symbol animation seems to be a bit off now. The "X" symbol is positioned a little bit to the left compared to the three hamburger slices:
image
I don't think that it already was like this before your latest pushes.

mgerhold (approved):

PR #16, opened 2024-09-06
Register

PurpurTentakel97:


src/routes/register/+page.svelte
```diff
@@ -0,0 +1,349 @@
...
+
+	function startTimerEmail() {
+		if (timerEmail !== null) {
+			clearTimeout(timerEmail);
+			timerEmail = null;
+		}
+		timerEmail = setTimeout(() => {
+			onEmailChanged();
+		}, 1000);
+	}
```
mgerhold, line 62: Could you extract the common code into a separate function (e.g. `startTimer()`) that takes the timer as parameter?

src/routes/register/+page.svelte
```diff
@@ -0,0 +1,349 @@
...
+		const reset = () => {
+			usernameMessage = '';
+		};
+
+		if (username.trim() === '') {
+			reset();
+			return;
+		}
+
+		const response = await fetch('api/account/username/exists?username=' + username);
```
mgerhold, line 78: Shouldn't the username also be trimmed here?
```suggestion
		const response = await fetch('api/account/username/exists?username=' + username.trim());
```

src/routes/register/+page.svelte
```diff
@@ -0,0 +1,349 @@
...
+		try {
+			data = await response.json();
+		} catch (error) {
+			console.error('error while parsing username json', error);
+			reset();
+			return;
+		}
+
+		if (data.exists) {
+			usernameMessage = 'Der Name  "' + username + '" ist bereits vergeben.';
```
mgerhold, line 94: Also here?
```suggestion
			usernameMessage = 'Der Name  "' + username.trim() + '" ist bereits vergeben.';
```

src/routes/register/+page.svelte
```diff
@@ -0,0 +1,349 @@
...
+			return;
+		}
+
+		reset();
+	}
+
+	async function onEmailChanged() {
```
mgerhold, line 101: This function is almost the same as the previous one. Maybe you could merge them into a single parameterized function.

src/routes/register/+page.svelte
```diff
@@ -0,0 +1,349 @@
...
+			return;
+		}
+
+		if (password_2.trim() === '') {
+			reset();
+			return;
+		}
+
+		if (!samePassword()) {
+			passwordMessage = 'Die Passwörter stimmen nicht überein.';
```
mgerhold, line 150: When registering, the error message "Die Passwörter stimmen nicht überein." appears while typing the repetition of the password. I find this a bit annoying. Should we maybe change it so that the message only appears if the repeated password is at least as long as the original password?

src/routes/register/+page.svelte
```diff
@@ -0,0 +1,349 @@
...
+		{:else}
+			<Headline>Registriert</Headline>
+			<div class="width-wrapper-registered">
+				<HorizontalLine />
+				<Spacer --height="7rem" />
+				<Textline --text-align="center">Deine Registrierung war erfolgreich.</Textline>
+				<Spacer --height="2rem" />
+				<Textblock --text-align="center">
+					Du bekommst jetzt eine Mail. Bestätige deine Mailadresse und du kannst dich anmelden.
+					Weitere Infos findest du nach dem anmelden.
```
mgerhold, line 273:
```suggestion
					Du bekommst jetzt eine Mail. Bestätige deine Mailadresse, um dich anmelden zu können.
					Weitere Infos findest du nach dem Anmelden.
```

src/routes/year/[year]/+page.js
```diff
@@ -0,0 +1,11 @@
...
+    const loggedInResult = await fetch('/api/account');
+    const loggedIn = loggedInResult.ok;
+
+    const currentYear = 2025; // TODO: load from database
+
+    return {
+        currentYear : currentYear,
+        loggedIn : loggedIn
+    }
+}
```
Comment (mgerhold) [inline src/routes/year/[year]/+page.js:11]: Missing newline

src/routes/year/[year]/year.svelte
```diff
@@ -1,546 +1,548 @@
...
 		['Ablaufplan', '#Shedule'],
-		['Login', '/login']
+		['Anmelden', '/login']
+	];
+	const headerMenuLoggedIn = [
+		['Start', '#top'],
+		['Vortragende', '#Speaker'],
+		['Partner', '#Sponsors'],
+		['Ablaufplan', '#Shedule'],
+		['Intern', '/backend'],
```
Comment (mgerhold) [inline src/routes/year/[year]/year.svelte:22]: Instead of showing "Intern" as a regular navigation item, what do you think of using a simple pictogram? Like this for example (as seen on JetBrains' website):
image
When clicking on it, a context menu could appear with multiple items like "Mein Profil", "Meine Talks", "Abmelden", etc.

static/Style.css
```diff
@@ -31,4 +44,4 @@ body {
 body {
     background-color: var(--background-color);
     transition: background-color var(--transition-duration);
-}
+}
```
mgerhold, line 47: Missing newline

mgerhold (commented):
A few things I noticed:
- When you're logged in and the main page reloads, it takes little moment until "Intern" appears. I understand that you have to make an API call to find out whether you're logged in or not. But would it be possible to do this *before* rendering the page? Maybe Svelte has something built-in to make that possible without too much effort.
P.S.: You said that the current implementation should already do this, but it doesn't work for me using Chrome.
- For the password repetition, I'd suggest changing the label text to "Passwort wiederholen:" instead of having "Passwort:" twice.
- You have a typo in the folder name "messa[n]ges"

src/elements/input/button.svelte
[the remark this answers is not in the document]
PurpurTentakel97: I tried to change that.
I did not change that in the mobile mode because there is no hover. I dont now if I should change that there.

src/elements/messanges/errorMessage.svelte
```diff
@@ -0,0 +1,35 @@
...
+	export let message = '';
+</script>
+
+{#if message !== ''}
+	<p transition:fade={{ duration: 300 }}>{message}</p>
+{/if}
+
+<style>
+	p {
+		background-color: red;
```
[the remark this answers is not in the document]
PurpurTentakel97: I changed it so some orange and introduced a CSS-variable for it.

src/routes/impressum/+page.svelte
```diff
@@ -1 +1,28 @@
...
+	import Headline from 'elements/text/headline.svelte';
+	import TextLine from 'elements/text/textline.svelte';
+	import Spacer from 'elements/spacer/spacer.svelte';
+	import Button from 'elements/input/button.svelte';
+</script>
+
+<div class="wrapper">
+	<Headline>Impressum</Headline>
+	<Spacer --height="3rem" />
+	<TextLine --text-align="center">TODO: Hier kommt das Beste Impressum hin.</TextLine>
```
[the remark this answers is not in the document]
PurpurTentakel97: I would do this after the merge in a new branch because I do that pressy easy without any merge conflicts.

src/routes/register/+page.svelte
PurpurTentakel97: yeah that worked with a ref workaround. youll see what i mean.

src/routes/year/[year]/year.svelte
Comment (PurpurTentakel97) [reply src/routes/year/[year]/year.svelte:23]: We should discuss that. There are basically two option.
fist ist to navigate into the backend an display a full navigation there. 
second is to display some navigation inside the "normal" page.
what do you prefer. And is that a topic for the registration and login branch or should be add an issue and merge this branch for now?

src/routes/register/+page.svelte
PurpurTentakel97: Im not able to pass the message values as a ref.

In Ts all primitive values are by value.
I could wrap it in an object so it is by ref.
But because i need to wrap that in html in a Lambda, so I can provide the values, is seems to be by value again.

So im not able to update the error message. 😓

PurpurTentakel97: But I really dont get why the workaround is working for the timer but not for a string.

mgerhold: Maybe strings are immutable and all changes will automatically create a new string instead of changing the existing one? (just a guess, because many languages do it like this)

src/routes/register/+page.svelte
mgerhold: Did it even have a timeout before? I can't remember. However, I think the current timeout is too long. It takes too much time until the error message appears. Also, another problem is that if I correct the error, it takes a long time for the error message to disappear. That can be confusing.
I don't really know if the timeout is the right way here. At least not in the general case. I'm thinking of a solution that depends on the actual input, e.g.:
- Is the repeated password shorter than the original password? → Use a timeout (~ 1 sec?)
- Is it as long as or even longer than the original? → Zero timeout, show message immediately
- Does it match the original password? → Hide all error messages immediately

Of course, I don't know how difficult this is to build.

src/routes/year/[year]/year.svelte
Comment (mgerhold) [reply src/routes/year/[year]/year.svelte:23]: Yes, this should not be part of the current pull request, you're right. On the question what I prefer: I just think that a small pictogram doesn't take up as much space as the additional navigation item. But regarding where to put the "backend" navigation (either main-page or on the sub-page), either way would be okay for me.

src/routes/register/+page.svelte
```diff
@@ -75,13 +72,15 @@
 			return;
 		}
 
-		const response = await fetch('api/account/username/exists?username=' + username);
+		const response: Response = await fetch(
+			'api/account/username/exists?username=' + username.trim()
+		);
 		if (!response.ok) {
 			usernameMessage = 'Fehler beim überprüfen des Namen. Fehlercode: ' + response.status;
```
mgerhold, line 79: I don't think we should show this to the user. Maybe the rate limiting (not yet implemented) has just rejected the request. Showing this to the user can be confusing. Maybe just ignore the error?

src/routes/register/+page.svelte
```diff
@@ -154,29 +153,29 @@
...
 		if (!response.ok) {
-			const entries = async (r) => {
-				const t = await r.text();
-				const j = JSON.parse(t);
+			const entries = async (response: Response): Promise<string[]> => {
+				const text: string = await response.text();
+				const j: { [key: string]: string } = JSON.parse(text);
 				return Object.values(j);
 			};
 			errorMessage = await entries(response);
```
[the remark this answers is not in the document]
PurpurTentakel97: That is correct.
That is because they result from your response.
They will clear by the text API call.
Because right now I can not know wich error message belongs to wich input line.
So I can clear only all errors or none.
I decided None because otherwise you would need to click register again to see the error messages for the other input lines again. 

What do you think?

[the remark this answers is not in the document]
mgerhold: I think this is a good argument to change the backend to that it returns error codes instead of messages. I mean, yes, you'll have to "translate" those to human-readable text then, but at least your code knows what the error means.

src/elements/text/subheadline.svelte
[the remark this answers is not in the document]
PurpurTentakel97: I did rename that and I was sure that git will commit it because of:
```
[core]
	ignorecase = false
```

But it did not.
I manually renamed it in gut now. 😅

src/helper/loggedIn.ts
```diff
@@ -0,0 +1,37 @@
+import { redirect } from '@sveltejs/kit';
+import { loginStatus } from 'stores/auth';
+
+async function fetchLoginStatus(fetch: Function): Promise<boolean> {
+    console.log("fetcching");
```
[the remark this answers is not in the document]
PurpurTentakel97: removed

src/helper/loggedIn.ts
```diff
@@ -0,0 +1,37 @@
...
+        redirect(302, link);
+    }
+    return status;
+}
+
+export function querryLogin() {
+    loginStatus.set(true);
+}
+
+export function querryLogout() {
```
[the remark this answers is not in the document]
PurpurTentakel97: removed -> updateLoginStatus(fetch);

src/helper/loggedIn.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export async function getLoginStatusAndCheckRedirect(fetch: Function, link: string, expectedLogginStatus: boolean): Promise<boolean> {
+    const status = await getLoginStatus(fetch);
+    if (status !== expectedLogginStatus) {
+        redirect(302, link);
+    }
+    return status;
+}
+
+export function querryLogin() {
```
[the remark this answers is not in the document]
PurpurTentakel97: removed -> updateLoginStatus(fetch);

src/routes/register/+page.svelte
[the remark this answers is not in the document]
PurpurTentakel97: Ill make an issue for that.

src/helper/loggedIn.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (status === null) {
+        const fetchedStatus = await fetchLoginStatus(fetch);
+        loginStatus.set(fetchedStatus);
+        return fetchedStatus;
+    }
+
+    return status;
+}
+
+export async function getLoginStatusAndCheckRedirect(fetch: Function, link: string, expectedLogginStatus: boolean): Promise<boolean> {
```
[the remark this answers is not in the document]
PurpurTentakel97: done

src/routes/backend/+page.ts
```diff
@@ -0,0 +1,9 @@
+import type { LoadBackendPromise } from "types/types";
+
+import { getLoginStatusAndCheckRedirect } from "helper/loggedIn";
+
+export async function load({ fetch }: { fetch: typeof globalThis.fetch }): LoadBackendPromise {
+    await getLoginStatusAndCheckRedirect(fetch, '/login', true);
```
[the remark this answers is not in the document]
PurpurTentakel97: done

src/elements/text/subheadline.svelte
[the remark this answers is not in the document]
mgerhold: Ah, now I understand! This is the typical problem on Windows where file paths are case-insensitive (which is stupid IMHO 😅)

src/routes/register/+page.svelte
PurpurTentakel97: I can try that. Well see. But this will result in a custom function. Later on will be the general call.

src/routes/register/+page.svelte
PurpurTentakel97: done

src/routes/register/+page.svelte
```diff
@@ -213,40 +212,48 @@
...
+						bind:value={password_2}
+						on:input={() => {
+							startTimer(timerPasswordRef, onPasswordChanged, 4000);
+						}}
 					/>
 
 					<div class="password-list-wrapper">
 						<Spacer --height="3rem" />
-						<Textline>Dein Password sollte folgendes enthalten:</Textline>
+						<TextLine>Dein Password sollte folgendes enthalten:</TextLine>
```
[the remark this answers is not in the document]
PurpurTentakel97: done

src/routes/year/[year]/year.svelte
```diff
@@ -1,33 +1,34 @@
-<script>
+<script lang="ts">
 	import Header from 'elements/header.svelte';
 	import Footer from 'elements/footer.svelte';
 	import SpacerHeader from 'elements/spacer/spacerHeader.svelte';
 
-	export let displayedYear;
-	export let currentYear;
-	export let loggedIn;
+	export let displayedYear: Number;
```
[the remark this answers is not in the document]
Comment (PurpurTentakel97) [reply src/routes/year/[year]/year.svelte:6]: I did.
This was right after I switched to TS.
The small versions are the primitive types.
The camel ones are some class.

src/routes/register/+page.svelte
PurpurTentakel97: I now have a function that sets the timer time based on the password length. What do you thin?

src/routes/register/+page.svelte
PurpurTentakel97: I would like to make an issue for this and mark it as bug.
Im not able to fix that right now.

src/routes/register/+page.svelte
mgerhold: I think it's way better than before👍 But, if the second password is not shorter than the first one, why do you use a timeout of 200 ms? You could display the error immediately then because you are absolutely sure that the user has done something wrong. So why still wait?

src/routes/register/+page.svelte
mgerhold: I'm fine with that 👍

src/routes/login/+page.svelte
```diff
@@ -10,44 +12,48 @@
...
 	];
 
-	let usernameOrEmail = '';
-	let password = '';
+	let usernameOrEmail: string = '';
+	let password: string = '';
 
-	let errorMessage = '';
+	const loggedInMessage = data.loggedIn ? 'Du bist bereits angemeldet.' : '';
+	const displayLoginMessage = data.showLoginMessage ? 'Du musst dich zunächst anmelden' : '';
```
mgerhold, line 33:
```suggestion
	const displayLoginMessage = data.showLoginMessage ? 'Du musst dich zunächst anmelden.' : '';
```

src/routes/login/+page.svelte
mgerhold, line 1: One thing we talked about just comes to my mind: We wanted a link on the login-page that says something like "Neues Konto erstellen" to make it easier to get from the login page to the registration page.
This can also be done in a separate PR, but we should not forget that.

src/routes/login/+page.svelte
PurpurTentakel97: done

mgerhold (approved):

PR #33, opened 2024-09-14
18 impressum

PurpurTentakel97:
close #18 
also see #31

PR #34, opened 2024-09-14
32 async prefix

### LINKED ISSUE #32: Async Prefix
Add a `Async` Prefix to all Async functions.

PurpurTentakel97:
close #32 
Prefix has turned to an Postfix. LUL

blocked by #33


mgerhold (approved):
I'm not really sure if this change in naming follows common practice and/or is a good thing in general (I just don't know enough about JS or TS), but I don't want to prevent you doing this.

PR #37, opened 2024-09-16
20 convert errors into codes

### LINKED ISSUE #20: convert errors into codes
use codes to return the error into the frontend.
That way I can use this to control the error messages.

PurpurTentakel97:
close #20 

chat pointed out that you should not point out if the username / mail is unknown or not.
so I lookup the same message for username / mail and invalid password.

### DIFF
--- a/src/lookup/registerLookup.ts
+++ b/src/lookup/registerLookup.ts
@@ -4,25 +4,25 @@ type RegisterDict = {
 
 const lookup: RegisterDict = {
     // username
-    "The username field is required.": "Server: Das Feld 'Name' ist erforderlich.",
-    "The username field must be at least 3 characters in length.": "Server: Der Name muss mindestens aus 3 Zeichen bestehen.",
-    "The username field cannot exceed 30 characters in length.": "Server: Der Name darf maximal aus 30 Zeichen bestehen.",
+    "The username field is required.": "Das Feld 'Name' ist erforderlich.",
+    "The username field must be at least 3 characters in length.": "Der Name muss mindestens aus 3 Zeichen bestehen.",
+    "The username field cannot exceed 30 characters in length.": "Der Name darf maximal aus 30 Zeichen bestehen.",
 
     // mail
-    "The email field is required.": "Server: Das Feld 'E-Mail' ist erforderlich.",
-    "The email field must contain a valid email address.": "Server: Das E-Mail-Feld muss eine gültige E-Mail-Adresse enthalten",
+    "The email field is required.": "Das Feld 'E-Mail' ist erforderlich.",
+    "The email field must contain a valid email address.": "Das E-Mail-Feld muss eine gültige E-Mail-Adresse enthalten",
     
     // username or mail
     "Username or email already taken" : "Name oder E-Mail sind bereits vergeben",
 
     // password
-    "The password field is required.": "Server: Das Feld 'Passwort' ist erforderlich.",
-    "PASSWORD_MISSING_FIELD": "Server: Das Feld 'Passwort' ist erforderlich.",
-    "PASSWORD_TOO_SHORT": "Server: Dein Passwort ist zu kurz. Mindestanzahl: 8 Zeichen",
-    "PASSWORD_MISSING_LOWERCASE_LETTER": "Server: Deinem Passwort fehlt ein kleiner Buchstabe.",
-    "PASSWORD_MISSING_UPPERCASE_LETTER": "Server: Deinem Passwort fehlt ein großer Buchstabe.",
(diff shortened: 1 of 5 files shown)


src/lookup/registerLookup.ts
```diff
@@ -0,0 +1,34 @@
+type RegisterDict = {
+    [key: string]: string;
+};
+
+const lookup: RegisterDict = {
+    // username
+    "The username field is required.": "Server: Das Feld 'Namen' ist erforderlich.",
```
mgerhold, line 7:
```suggestion
    "The username field is required.": "Server: Das Feld 'Name' ist erforderlich.",
```

src/lookup/registerLookup.ts
```diff
@@ -0,0 +1,34 @@
...
+    // username or mail
+    "Username or email already taken" : "Name oder E-Mail sind bereits vergeben",
+
+    // password
+    "The password field is required.": "Server: Das Feld 'Passwort' ist erforderlich.",
+    "PASSWORD_MISSING_FIELD": "Server: Das Feld 'Passwort' ist erforderlich.",
+    "PASSWORD_TOO_SHORT": "Server: Dein Passwort ist zu kurz. Mindestanzahl: 8 Zeichen",
+    "PASSWORD_MISSING_LOWERCASE_LETTER": "Server: Deinem Passwort fehlt ein kleiner Buchstabe.",
+    "PASSWORD_MISSING_UPPERCASE_LETTER": "Server: Deinem Passwort fehlt ein großer Buchstabe.",
+    "PASSWORD_MISSING_NUMBER": "Server; Deinem Passwort fehlt eine Zahl.",
```
mgerhold, line 24:
```suggestion
    "PASSWORD_MISSING_NUMBER": "Server: Deinem Passwort fehlt eine Zahl.",
```

src/lookup/registerLookup.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    // mail
+    "The email field is required.": "Server: Das Feld 'E-Mail' ist erforderlich.",
+    "The email field must contain a valid email address.": "Server: Das E-Mail-Feld muss eine gültige E-Mail-Adresse enthalten",
+    
+    // username or mail
+    "Username or email already taken" : "Name oder E-Mail sind bereits vergeben",
+
+    // password
+    "The password field is required.": "Server: Das Feld 'Passwort' ist erforderlich.",
```
mgerhold, line 19: Why do you prefix those messages with `Server: `? Is this important for the user?

PurpurTentakel97: no it is not I think.
in the bast case the user does not see this messages at all. 
Bit for development it is interesting because it decently is a case I did not cover.

PurpurTentakel97:
> What's your opinion?

I think your right.
Should I change the message then?

src/lookup/registerLookup.ts
mgerhold: Yes, but if it's a "for development only" message, shouldn't it be in English? And maybe it should only be logged to the console instead of shown to the user.

mgerhold:
> I think your right. Should I change the message then?

I think it's okay to just tell the user the facts (e.g. "username already taken", "email already taken" etc.)

src/lookup/registerLookup.ts
PurpurTentakel97: I think I should make the lookup and display it in German. But I can extra make an error log with the englisch text. 
But then I would remove the `Server:`

mgerhold: At the moment I think that we should remove those "fallback" cases (regarding the user-facing messages). If everything works fine, the user won't ever see those messages. That means: If the user sees the message, that's a bug on our end. I think we should just remove that and only log it to the console.
But I don't know if this is the right thing to do.

PurpurTentakel97: Do you really think?
The User will not understand why its not working.
But he could when I would display the message.
Than they can change their entries.

In best case it will never happen.  Thats right.
But I think the user should have the chance to correct its values.

PurpurTentakel97: The solution as it is in this PR will print  by every negative API call.
In the next PR there will be return that prevents the API call when the frontend has already detected an error.

mgerhold: > [...] The User will not understand why its not working. But he could when I would display the message. Than they can change their entries.

I think the error message could just be "Interner Fehler", "Unbekannter Fehler", "Ungültige Daten", or something like that. There's no need to use the "internal" error message.

mgerhold (approved):

PR #52, opened 2024-09-30
40 parse json

### LINKED ISSUE #39: email validation
fix email validation in register validation.

### LINKED ISSUE #40: parse json
Find a better way to parse a json without that much boiler plate code.

PurpurTentakel97:
close #40 
close #39 

This adds a plugin that validate Incoming  json and email.
I was not able to get the default tlds to work so I defined a bunch of tip-level-somains to validate against.
I hope that will work.

### DIFF
--- a/src/helper/parseJson.ts
+++ b/src/helper/parseJson.ts
@@ -12,8 +12,6 @@ export async function parseProvidedJsonAsync<T>(response: Response, scheme: ZodT
         console.error(validated.error.errors);
         console.log("Validation-Original:", type);
 
-        return type;
-
     } catch (error) {
         console.error('Error occurred during fetch or validation:', error);
     }
--- a/src/helper/parseJson.ts
+++ b/src/helper/parseJson.ts
@@ -1,0 +1,14 @@
+import type Joi from "joi";
+
+export async function parseProvidingJson<T>(response: Response, scheme: Joi.Schema<T>): Promise<T> {
+    const type: T = await response.json();
+    const { error } = scheme.validate(type);
+
+    if (error) {
+        console.error(error);
+        console.log("Validation-Original:", error._original);
+        console.log("Validation-Details:", error.details);
+    }
+
+    return type;
+}
--- a/src/helper/parseJson.ts
+++ b/src/helper/parseJson.ts
@@ -1,6 +1,6 @@
 import type Joi from "joi";
 
-export async function parseProvidingJson<T>(response: Response, scheme: Joi.Schema<T>): Promise<T> {
+export async function parseProvidedJsonAsync<T>(response: Response, scheme: Joi.Schema<T>): Promise<T> {
     const type: T = await response.json();
     const { error } = scheme.validate(type);
 
--- a/src/routes/register/validation.ts
+++ b/src/routes/register/validation.ts
@@ -1,6 +1,6 @@
 import { apiUrl } from "helper/links";
-import Joi, { options } from "joi";
-import { parseProvidingJson } from "helper/parseJson";
+import Joi from "joi";
+import { parseProvidedJsonAsync } from "helper/parseJson";
 
 const tlds: string[] = [
     'com', 'org', 'net', 'info', 'web',
@@ -38,7 +38,7 @@ export async function onUsernameChangedAsync(username: string, fetch: Function):
         return;
     }
 
-    const data: Data = await parseProvidingJson<Data>(response, dataScheme);
+    const data: Data = await parseProvidedJsonAsync<Data>(response, dataScheme);
 
     if (data.exists) {
         return 'Der Name  "' + trimmed + '" ist bereits vergeben.';
@@ -63,7 +63,7 @@ export async function onMailChangedAsync(mail: string, fetch: Function): Promise
         return;
     }
 
-    const data: Data = await parseProvidingJson<Data>(response, dataScheme);
+    const data: Data = await parseProvidedJsonAsync<Data>(response, dataScheme);
 
     if (data.exists) {
         return 'Die E-Mail "' + trimmed + '" wird bereits verwendet.';
--- a/src/routes/register/validation.ts
+++ b/src/routes/register/validation.ts
@@ -1,6 +1,13 @@
 import { apiUrl } from "helper/links";
-import Joi from "joi";
+import Joi, { options } from "joi";
 import { parseProvidingJson } from "helper/parseJson";
+
+const tlds: string[] = [
+    'com', 'org', 'net', 'info', 'web',
+    'us', 'uk', 'ca', 'de', 'fr', 'au', 'jp', 'cn', 'in', 'br', 'ru', 'it', 'es',
+    'mx', 'nl', 'se', 'ch', 'no', 'fi', 'hk', 'kr', 'sg', 'za', 'pt', 'dk',
+    'ie', 'co', 'cl', 'tw', 'ae',
+];
 
 type Data = {
     exists: boolean;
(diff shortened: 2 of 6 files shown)


src/helper/parseJson.ts
```diff
@@ -0,0 +1,14 @@
+import type Joi from "joi";
+
+export async function parseProvidedJsonAsync<T>(response: Response, scheme: Joi.Schema<T>): Promise<T> {
+    const type: T = await response.json();
```
PurpurTentakel97, line 4: Im not quite sure if I should use a try catch here.
What do you think.
That can only fail when the backend does not send valid json.
Maybe I do add it and return an empty T then?

mgerhold: I'd say: In general never trust data coming from the outside. So a `try` would be a good thing here, I think.

src/routes/register/validation.ts
```diff
@@ -1,4 +1,21 @@
...
+import { parseProvidedJsonAsync } from "helper/parseJson";
+
+const tlds: string[] = [
+    'com', 'org', 'net', 'info', 'web',
+    'us', 'uk', 'ca', 'de', 'fr', 'au', 'jp', 'cn', 'in', 'br', 'ru', 'it', 'es',
+    'mx', 'nl', 'se', 'ch', 'no', 'fi', 'hk', 'kr', 'sg', 'za', 'pt', 'dk',
+    'ie', 'co', 'cl', 'tw', 'ae',
+];
+
+type Data = {
```
mgerhold, line 12:
```suggestion
type UsernameExistsResponse = {
```

src/routes/register/validation.ts
```diff
@@ -41,8 +52,8 @@ export async function onMailChangedAsync(mail: string, fetch: Function): Promise
         return 'Das Feld "E-Mail" ist leer.';
     }
 
-    const valid: boolean = /^[^\s@]+@[^\s@]+\.[^\s@]+$/.test(trimmed);
-    if (!valid) {
+    const { error } = Joi.string().email({ tlds: { allow: tlds } }).required().validate(trimmed);
```
mgerhold, line 55: If we really want this, I think I'd have to extend the backend as well. But I don't know why we should restrict the top level domains.

PurpurTentakel97: The new lib is validating the top level domain automatically.
Is that okay?

src/routes/register/validation.ts
```diff
@@ -21,16 +28,14 @@ export async function onUsernameChangedAsync(username: string, fetch: Function):
         return;
     }
 
-    let data: { exists: boolean };
-    try {
-        data = await response.json();
-    } catch (error) {
-        console.error('error while parsing username json', error);
+    const data: ExistsResponseData | undefined = await parseProvidedJsonAsync<ExistsResponseData>(response, existsResponseDataScheme);
```
peddermaster2, line 31: Is it nessesary to add the full type here? It should be possible to infer that type from the function call.

src/routes/register/validation.ts
```diff
@@ -52,16 +57,14 @@ export async function onMailChangedAsync(mail: string, fetch: Function): Promise
         return;
     }
 
-    let data: { exists: boolean };
-    try {
-        data = await response.json();
-    } catch (error) {
-        console.error('error while parsing email json', error);
+    const data: ExistsResponseData | undefined = await parseProvidedJsonAsync<ExistsResponseData>(response, existsResponseDataScheme);
```
peddermaster2, line 60: Same as above.

src/routes/year/[year]/year.ts
```diff
@@ -18,10 +21,13 @@ export async function loadYearAsync(fetch: Function, year: number | undefined =
...
 
-    let yearData: Year;
-    if (yearDataResponse.ok) {
-        yearData = await yearDataResponse.json();
-    } else {
+    if (!yearDataResponse.ok) {
+        redirect(302, "/404");
+    }
+
+    const yearData: Year | undefined = await parseProvidedJsonAsync<Year>(yearDataResponse, yearScheme);
```
Comment (peddermaster2) [inline src/routes/year/[year]/year.ts:28]: Same as in `src/routes/register/validation.ts:31`

src/routes/register/validation.ts
PurpurTentakel97: Im not quite sure what to do here.
Should I remove the Type?

I normally like to write the types down to see what Type should be there.
But you're right. Because auf the return type of the function one can remove it. 

what do you think @mgerhold ?

mgerhold: If the type can be infered, I'd suggest leaving it out. Only exception: Function signatures (those should always have type annotations). At least this is how I do it in Python...
I'm no TS expert

src/helper/parseJson.ts
```diff
@@ -0,0 +1,20 @@
...
+        const validated = scheme.safeParse(type);
+
+        if (validated.success) {
+            return validated.data;
+        }
+
+        console.error(validated.error.errors);
+        console.log("Validation-Original:", type);
+
+        return type;
```
mgerhold, line 15: Will `type` be `undefined` in this case?

src/routes/register/validation.ts
```diff
@@ -52,16 +57,14 @@ export async function onMailChangedAsync(mail: string, fetch: Function): Promise
...
+    if (!data) {
+        console.log('Error while checking mail');
```
mgerhold, line 62:
```suggestion
        console.error('Error while checking mail');
```

src/routes/register/validation.ts
```diff
@@ -21,16 +28,14 @@ export async function onUsernameChangedAsync(username: string, fetch: Function):
...
     }
 
-    let data: { exists: boolean };
-    try {
-        data = await response.json();
-    } catch (error) {
-        console.error('error while parsing username json', error);
+    const data: ExistsResponseData | undefined = await parseProvidedJsonAsync<ExistsResponseData>(response, existsResponseDataScheme);
+    if (!data) {
+        console.log('Error while checking username');
```
mgerhold, line 33:
```suggestion
        console.error('Error while checking username');
```

src/routes/register/validation.ts
mgerhold: The new lib seems to be quite nice compared to the old one 👍 I think this is good!

src/helper/parseJson.ts
PurpurTentakel97: In my understanding the type will be not undefined when there is no exception.
That can be undefined are the single fields within the type.
Im not quite sure if that is a problem. I normal only display the data. 
So in worst case there will be displayed undefined on the website. 

But if you think its to risky I can just return undefined in that case.
That is maybe easier to debug anyway.
What do you think?

src/routes/register/validation.ts
PurpurTentakel97: The Parse Function is already printing an error here.
So I see this more like a hint.
Would you make it an error anyway?

src/helper/parseJson.ts
mgerhold: I just expected that this function would return undefined if an error occured. But instead you seem to return a `T` object which may have invalid contents. Maybe then better return an undefined value? I'm not sure.

src/routes/register/validation.ts
mgerhold: Ah okay, I see. However, I think an error would still be more appropriate here 🤔

peddermaster2 (approved):

mgerhold (approved):

PR #56, opened 2024-10-08
Sponsors

PurpurTentakel97:
Adds social media links from backend.
Adds team members from backend.
Adds Sponsor links.

### DIFF
--- a/src/elements/image/iconLink.svelte
+++ b/src/elements/image/iconLink.svelte
@@ -1,11 +1,11 @@
 <script lang="ts">
 	import Icon from './icon.svelte';
-	import type { IconLink } from 'types/provideTypes';
-	export let linkType: IconLink;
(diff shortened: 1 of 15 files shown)


mgerhold (commented):
That's really cool!
Only one thing: I'd like it more if the sponsor and partner images would be regular images, not "buttons". Can we get rid of this hover effect?
image
Or do you think it's better this way?

PurpurTentakel97:
That would be like on the old side.
Would that be okay for you?

For me there is nothing happening, wenn hovering and clicking.
So I'll tried to bring some Interaktion to that.

You can look at the old page when you want a reference for no effect.

mgerhold:
I like the look and feel of the old page regarding the sponsors, yes.

mgerhold (approved):

PR #58, opened 2024-10-09
54 fix global

PurpurTentakel97:
close #54 
close #57 

1.
This PR makes sure `:global()` does not leak outside of a component.
To be able to style a Component from the outside I usually provide the class as a string and style it with `:global()`.
By using `:global()` svelte no longer applies the random CSS classes to generate the "capsular CSS".
By selecting an other CSS class before `:global()` the other class has this extra random CSS class and only when `:global()` is a child (in any kind, not only direct child) it gets selected.
So there is a kind of capsular again. 
Does that makes sense to you?

2.
it uses sponsor and media partner data from the backend. 

3. 
It introduces a global `hover-move` CSS class to reduce duplicate code.

PR #59, opened 2024-10-09
Hover

PurpurTentakel97:
blocked by #58 (review that first, here is only the last commit interesting for you)

An other try to hover for sponsors and media partners.
Its not that heavy.
What do you think?


mgerhold (dismissed):
I think this is okay 👍

PurpurTentakel97:
reopen soon because the approval was gone.

PR #61, opened 2024-10-10
60 use svelte for 404 page

### LINKED ISSUE #60: use svelte for 404 page
add a `+error.svelte` into the `routes` directory.
replace the current 404 page. 

example code:
```svelte
<script>
   export let error;
   export let status;
</script>

{#if status === 404}
   <h1>Page Not Found</h1>
   <p>The page you're looking for doesn't exist.</p>
{:else}
   <h1>{status}</h1>
   <p>{error.message}</p>
{/if}

<a href="/">Go back home</a>
```

Use `throw error(404, msg);` instead of `redirect(404, link);` to navigate on purpose to the error side.

Consider a contact to provide a possibility for error reports.

PurpurTentakel97:
close #60 

adds a small zoom effect for sponsors ans media partners (already approved)

add a "real" error page that gets displayed for all svelte errors.
also throw errors instead of redirecting to custom 404 page.


src/routes/+error.svelte
```diff
@@ -0,0 +1,60 @@
...
+
+	function onClick() {
+		goto('/');
+	}
+</script>
+
+<Header menu={Menu.header} />
+<div class="wrapper page-404">
+	<div class="content">
+		<HeadlinePage classes="border">{$page.status}</HeadlinePage>
```
mgerhold, line 21: I think we should use the status code and perform a lookup to show a German error message, e.g. "Uuups! Die Seite konnte nicht gefunden werden 👻"

src/routes/+error.svelte
```diff
@@ -0,0 +1,60 @@
...
+		{#if $page.error}
+			<SubHeadline classes="subheadline">Fehler: {$page.error.message}</SubHeadline>
+		{/if}
+
+		<Button
+			classes="text button"
+			ariaLabel="Klicke um zur Hauptseite zu navigieren"
+			on:click={onClick}
+		>
+			Hauptseite
```
mgerhold, line 31: On all other pages, you call this "Start", but here you call it "Hauptseite". I'm not sure which I like more (I always thought that "Start" is a bit weird), but at least it should be harmonized across all pages.

src/routes/year/[year]/+page.ts
```diff
@@ -1,12 +1,12 @@
...
-import { redirect } from "@sveltejs/kit";
+import { error } from "@sveltejs/kit";
 
 import { loadYearAsync } from "./year";
 
 export async function load({ fetch, params }: { fetch: typeof globalThis.fetch, params: { year: string } }): LoadYearPromise {
     const year: number = Number(params.year);
     if (Number.isNaN(year)) {
-        redirect(302, "/404");
+        throw error(404, `in der URL angegebenes Jahr "${params.year}" ist keine Zahl.`);
```
Comment (mgerhold) [inline src/routes/year/[year]/+page.ts:9]: I don't think we need error messages that are this specific. For the user this should just show that the page doesn't exist (404 with a general-purpose error message).

src/routes/year/[year]/year.ts
```diff
@@ -22,13 +22,13 @@ export async function loadYearAsync(fetch: Function, year: number | undefined =
...
     if (!yearDataResponse.ok) {
-        redirect(302, "/404");
+        throw error(400, `Serveranfrage für das Jahr ${year} nicht erfolgreich.`)
     }
 
     const yearData = await parseProvidedJsonAsync<Year>(yearDataResponse, yearScheme);
 
     if (!yearData) {
-        redirect(302, "/404");
+        throw error(400, `Unerwartete Daten für das Jahr ${year}.`)
```
Comment (mgerhold) [inline src/routes/year/[year]/year.ts:31]: Same here: I think those error messages are too specific to show them to the user. They are interesting for us (so maybe `console.error()` would be useful here?), but the user should just see a general 404 error message.

mgerhold (changes requested):

src/routes/+error.svelte
PurpurTentakel97: Would you only do that for 404?
The user will not see any other code or there is a serious issue.

What do you think?

src/routes/+error.svelte
PurpurTentakel97: So I turn it to "Hauptseite" then? :)

PurpurTentakel97:
Im thinking about a "contact us" `mail to` Link.
What do you think?

But I don't know what mail I should display.

src/routes/+error.svelte
mgerhold: Yes, then let's only do an error message for 404. We could also do another message for *any other* error, e.g. "Uuuups, irgendwas ist mächtig schief gelaufen, sorry! 🫨"

(the cringy emojis are very important Kappa)

src/routes/+error.svelte
mgerhold: Would "Hauptseite" fit into the header on all pages? Even on mobile? I don't know. On the other hand: Maybe we could remove it from the header altogether and only use the logo (in the upper left corner) as a link to the main page. What do you think?

mgerhold:
> Im thinking about a "contact us" `mail to` Link. What do you think?
> 
> But I don't know what mail I should display.

You mean on the error page? Wouldn't a general contact page be enough? I don't think that the error page needs that feature.

src/routes/+error.svelte
PurpurTentakel97: The logo is already a shortcut for that.

Right now it is fitting inside the header.
On mobile even better ;)

Check it out and let me know.
For me its okay as it is now. :)

mgerhold: But now it says "Top" instead of "Hauptseite" 🤔
image
Is this intended?

src/routes/+error.svelte
```diff
@@ -20,7 +32,15 @@
...
 		<HeadlinePage classes="border">{$page.status}</HeadlinePage>
 		{#if $page.error}
-			<SubHeadline classes="subheadline">Fehler: {$page.error.message}</SubHeadline>
+			{#if $page.status === 404}
+				<SubHeadline classes="subheadline">
+					Uuups! Die Seite konnte nicht gefunden werden 👻
+				</SubHeadline>
+			{:else}
+				<SubHeadline classes="subheadline">
+					Uuuups, irgendwas ist mächtig schief gelaufen, sorry! 🫨
```
mgerhold, line 41: Cringe! I like it! <3

mgerhold (changes requested):
As I wrote in the other comment, on the main page the nav bar now shows "Top" instead of "Hauptseite". The rest is good 👍

PurpurTentakel97:
You did not get that quite right.
"Top" was always "Top".
But I changed "Start" to "Hauptseite"

"Top" and "Hauptseite" does not do the same.
"Hauptseite" navigates to the main page of the current year.
"Top" scrolls to the top of the displayed year.
Therefore I would not name that the same.
When the current year and the displayed year is the same. Both do the same. But when they differ. "Hauptseite" loads an other page while "Top" does not.


But I see that the naming "Top" is not the yellow from the egg.
I don't like that ether.
But I don't know how else to name it. 😅

mgerhold:
Ah okay, now I get it. Then I will approve the PR for now. But I don't really like "Top" there. Maybe we should just show an arrow pointing upwards in a box, e.g. on the right side of the page. The box could be invisible when the user is already at the top and become visible when they scroll down.

mgerhold (approved):

Issue #41, opened 2024-09-22
social media icons


PurpurTentakel97:
try to add social media icons once again

PR #62, opened 2024-10-13
Schedule

PurpurTentakel97:
talks gets not displayed and slitted in dates and special talks.
The special talks are labelnd with "YouTube"
Also the popup inside the schedule is working.


src/lookup/tagColorLookup.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+
+const lookupTextColor: TagColorDict = {
+    "1": "#000",
+    "2": "#000",
+    "3": "#fff",
+    "4": "#fff",
+    "5": "#fff",
+    "6": "#000",
+}
```
mgerhold, line 21: Couldn't we use numbers here instead of strings?

src/lookup/tagColorLookup.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export function tagColorLookup(key: string): string {
```
mgerhold, line 23: Also here, I think this function could both take and return a number instead of a string, couldn't it?
(same for `tagTextColorLookup`, too)

src/lookup/tagColorLookup.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (key in lookupTagColor) {
+        return lookupTagColor[key];
+    }
+
+    return key;
```
mgerhold, line 28: What do you think about a wrap-around to ensure that the index will be valid?
```suggestion
    if (key < 0) {
        // throw error or simply return lookupTagColor[0]? Or some special "error" color?
    }
    return lookupTagColor[key % lookupTagColor.length];
```

src/routes/year/[year]/year.svelte
```diff
@@ -14,22 +12,56 @@
...
+
+		for (let talk of talks) {
+			if (!talk.is_special) {
+				continue;
+			}
+			toReturn.push(talk);
+		}
+
+		return toReturn;
+	}
```
Comment (mgerhold) [inline src/routes/year/[year]/year.svelte:64]: Should we also split the special talks by date? I don't know if we have/want to support YouTube talks for both days 🤔

mgerhold (commented):
I really like the visuals, looks very good 👍

src/routes/year/[year]/year.svelte
Comment (PurpurTentakel97) [reply src/routes/year/[year]/year.svelte:64]: What do you this this would look like?
something like this?
```
[
    [
         [
             // Normal Data day one
         ] , [
            // special Data day one
         ] ,
    ] , [
         [
             // Normal Data day two
         ] , [
            // special Data day two
         ] ,
    ],
]
```

And then I would ittertate in html.
In this case the special talks would be displayed right after die normal talks.
And if an array is empty the Schedule is doing nothing.
So just normal or just special talks would be perfectly fine.

src/lookup/tagColorLookup.ts
PurpurTentakel97: why you using the modulo there?
Than the color will "wrap around" right?

Wouldn't it be better to just throw an error?
Because the color code is not the expected than.

src/lookup/tagColorLookup.ts
PurpurTentakel97: I try

src/lookup/tagColorLookup.ts
PurpurTentakel97: I dont know why this even works right now.
The color index is a number by now.
There should be a type error. 🫤

src/routes/year/[year]/year.svelte
Comment (mgerhold) [reply src/routes/year/[year]/year.svelte:64]: Yes, that's also how I would imagine it. The only question is: Do we need this? Do we want this? I don't know, if this is a YAGNI situation

src/lookup/tagColorLookup.ts
mgerhold: Ah okay, that was a little mis-assumption on my end. In the admin dashboard, when we create a new tag, we will select a color there (based on the available colors of the frontend). So yes, you're right, there should never be an invalid color index (unless we reduce the number of indices later on, but that would not make any sense).
Okay, then you can just throw an error, I guess (instead of returning the key if it cannot be found).

src/lookup/tagColorLookup.ts
PurpurTentakel97: I leave that as a string because I need to convert it anyway.
CSS can only handle it as a string and then a have to convert it and do some string operations to match die CSS Hexa style.
`0xffeb3b` -> `"#ffeb3b"`

I Think its not worth it.

But I change the keys.

mgerhold: I think that's good, yes!

src/routes/year/[year]/year.svelte
Comment (PurpurTentakel97) [reply src/routes/year/[year]/year.svelte]: I changed that.
It should work but I can not check if the first day would actually display any talks because of missing data.
But when the loop works on the second day it will also work on all other days. 

It was not that hard and I think its even cleaner now. What do you think?

src/routes/year/[year]/year.svelte
```diff
@@ -31,36 +31,27 @@
 		personPopup = undefined;
 	}
 
-	function splitTalks(): Talk[][] {
+	function splitTalks(): Talk /* days */[] /* 0 = normal; 1 = special */[] /* talks */[] {
```
Comment (mgerhold) [inline src/routes/year/[year]/year.svelte:34]: Wouldn't it be good to use a data structure that has field names instead of using the comments here? It would also be more readable. Maybe write a dedicated class for this?

mgerhold (commented):
I created test data for a special talk on day one (see the PR [here](https://github.com/TechStreamConference/website-backend/pull/41). It looks perfectly fine 👍

I was wondering whether we should display the date of the bonus contents or not, but I think that it's good to show it when the list of regular talks gets longer and longer.

From looking at the website, I'd approve this right away. There's just this little comment regarding the code. But please decide for yourself if you want to change the code or not.

src/routes/year/[year]/year.svelte
Comment (PurpurTentakel97) [reply src/routes/year/[year]/year.svelte]: I switched to a type.
Because ts don't have any default constructors a need a function for that.
But I think its better now.

PurpurTentakel97:
Testdata working.

mgerhold (approved):

src/routes/year/[year]/year.svelte
Comment (mgerhold) [reply src/routes/year/[year]/year.svelte]: Thanks! I like it a lot more 👍

Issue #44, opened 2024-09-24
fontsize in person popup


PurpurTentakel97:
@mgerhold complained that the font size inside the speaker popup is to small.
Lets discuss that.

PR #64, opened 2024-10-19
display globals

PurpurTentakel97:
this uses the new globals fields from the backend and displays it onto the website.

### DIFF
--- a/src/elements/navigation/footer.svelte
+++ b/src/elements/navigation/footer.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-	import type { MenuItem } from 'types/provideTypes';
+	import type { Globals, MenuItem } from 'types/provideTypes';
 	import TextLine from 'elements/text/textLine.svelte';
 	import List from 'elements/list/list.svelte';
 	import ListElement from 'elements/list/listElement.svelte';
@@ -7,13 +7,13 @@
 	import Paragraph from 'elements/text/paragraph.svelte';
 	import LogoBig from 'elements/image/logoBig.svelte';
 
-	export let currentYear: number;
+	export let globals: Globals = { default_year: 0, footer_text: 'some text' }; // default values so that numbers can not fail because of undefiend
 	export let menu: MenuItem[];
 
+	const currentYear: number = new Date().getFullYear();
 	const startYear: number = 2024;
-
 	$: numbers = Array.from(
-		{ length: currentYear + 1 - startYear },
+		{ length: globals.default_year + 1 - startYear },
 		(_, i) => startYear + i
 	);
 </script>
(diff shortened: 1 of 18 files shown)


src/elements/navigation/footer.svelte
```diff
@@ -1,19 +1,19 @@
...
+	import type { Globals, MenuItem } from 'types/provideTypes';
 	import TextLine from 'elements/text/textLine.svelte';
 	import List from 'elements/list/list.svelte';
 	import ListElement from 'elements/list/listElement.svelte';
 	import Link from 'elements/text/link.svelte';
 	import Paragraph from 'elements/text/paragraph.svelte';
 	import LogoBig from 'elements/image/logoBig.svelte';
 
-	export let currentYear: number;
+	export let globals: Globals = { default_year: 0, footer_text: 'some text' }; // default values so that numbers can not fail because of undefiend
```
mgerhold, line 10:
```suggestion
	export let globals: Globals = { default_year: 0, footer_text: 'some text' }; // default values so that numbers can not fail because of undefined
```

src/elements/navigation/footer.svelte
```diff
@@ -1,19 +1,19 @@
...
 	export let menu: MenuItem[];
 
+	const currentYear: number = new Date().getFullYear();
 	const startYear: number = 2024;
-
 	$: numbers = Array.from(
-		{ length: currentYear + 1 - startYear },
+		{ length: globals.default_year + 1 - startYear },
 		(_, i) => startYear + i
 	);
```
mgerhold, line 18: Are those the years that get displayed in the footer? Is there something wrong?
image
We should only ever show years which have a (published) event, don't you think?

src/routes/+error.svelte
```diff
@@ -4,22 +4,45 @@
...
+	import { parseProvidedJsonAsync } from 'helper/parseJson';
 
 	onMount(async (): Promise<void> => {
-		data = defaultCurrentYear; // API call when endpoint is implemented
-		if (data) {
+		// dont use all helper functions because this page is a fallback
+		const handleFail = () => {
+			data = {
+				default_year: 0,
+				footer_text: 'some Fail'
```
mgerhold, line 24: Is this what the user will see on the error page of the globals cannot be retrieved? If so, maybe use some more reasonable data here, e.g. the current year and an empty string.

src/routes/+error.svelte
```diff
@@ -4,22 +4,45 @@
...
 	import HeadlinePage from 'elements/text/headlinePage.svelte';
 	import SubHeadline from 'elements/text/subHeadline.svelte';
+	import { apiUrl } from 'helper/links';
+	import { globalsScheme, type Globals } from 'types/provideTypes';
+	import { parseProvidedJsonAsync } from 'helper/parseJson';
 
 	onMount(async (): Promise<void> => {
-		data = defaultCurrentYear; // API call when endpoint is implemented
-		if (data) {
+		// dont use all helper functions because this page is a fallback
```
mgerhold, line 20: What do you mean by this? Is it okay to use helper functions, but not all of them? Or do you mean to not use *any* of the helper functions?
Also, there's a little typo: `dont` ⇒ `don't`

src/routes/imprint/+page.ts
```diff
@@ -1,13 +1,20 @@
 import type { LoadImprintPromise } from "types/loadTypes";
+import type { Globals } from "types/provideTypes";
+import { apiUrl } from "helper/links";
 import { getLoginStatusAsync } from "helper/loggedIn";
-import { defaultCurrentYear } from "delete/toDelete";
```
mgerhold, line 3: Have you also deleted the `delete` file?

src/elements/navigation/footer.svelte
PurpurTentakel97: How did you get that many years in there?
Normally there should be only an many years until the `default_year` is hit.

So right now only until `2024`.
But yes. Maybe the `globals` endpoint should provide an array and I replace this.

mgerhold: I don't know how I managed to do this ¯\\\_(ツ)_/¯
I tried it again and now it looks fine.

But a agree that it would be good to provide a list of the relevant years. I implemented this here: https://github.com/TechStreamConference/website-backend/issues/44

PR #73, opened 2024-10-25
Propz

### LINKED ISSUE #66: prevent routing while changes in dashboard

### LINKED ISSUE #68: Tap line in Dashboard not aligning

PurpurTentakel97:
I use a feature branch now to make the single PRs a little smaller

### new
- steaky header instead of fixed header
- prewrap some text so it breaks on `\n` (there are basically two methods. Frist with prewrap. Second with a function that turn `\n` into `<br/>`. both is fine I think. What do you like better?)
- aria labels are no longer hidden behind a `aria-hidden="true"`
- 404 page logs error
- improved z-index for header and popups.
- prevent navigating while unsaved changes.
- reset original data when navigated with unsaved changes. (or better just don't save the changed data back zu to original data.)
- round edges for tabs

### issues
- close #68 
- close #66


PurpurTentakel97:
open against main

PR #74, opened 2024-10-26
Propz

### LINKED ISSUE #65: use year array to display available years

### LINKED ISSUE #66: prevent routing while changes in dashboard

### LINKED ISSUE #68: Tap line in Dashboard not aligning

PurpurTentakel97:
I use a feature branch now to make the single PRs a little smaller

### new
- steaky header instead of fixed header
- prewrap some text so it breaks on `\n` (there are basically two methods. Frist with prewrap. Second with a function that turn `\n` into `<br/>`. both is fine I think. What do you like better?)
- aria labels are no longer hidden behind a `aria-hidden="true"`
- 404 page logs error
- improved z-index for header and popups.
- prevent navigating while unsaved changes.
- reset original data when navigated with unsaved changes. (or better just don't save the changed data back zu to original data.)
- round edges for tabs
- possible years from database are now in footer and dashboard

### issues
- close #68 
- close #66 
- close #65


src/elements/navigation/unsavedChangesPopup.svelte
mgerhold, line 1: I think the popup should have a background. Depending on the position of the popup, it can be hard to read:
image

mgerhold (dismissed):
(👉ﾟヮﾟ)👉

src/elements/navigation/footer.svelte
```diff
@@ -6,16 +6,12 @@
 	import Link from 'elements/text/link.svelte';
 	import Paragraph from 'elements/text/paragraph.svelte';
 	import LogoBig from 'elements/image/logoBig.svelte';
+	import { date } from 'zod';
 
-	export let globals: Globals = { default_year: 0, footer_text: 'default text' }; // default values so that numbers can not fail because of undefined
+	export let globals: Globals = { default_year: 0, footer_text: 'default text', years_with_events: [] }; // default values so that numbers can not fail because of undefined
```
peddermaster2, line 11: The comment seems a bit outdated.

src/elements/input/dropDown.svelte
```diff
@@ -0,0 +1,49 @@
...
+	function handleKeydown(event: KeyboardEvent) {
+		if (event.ctrlKey && event.key === 'Enter') {
+			dispatch('submit');
+		}
+	}
+</script>
+
+<div class="drop-down-wrapper">
+	<Label {id}>{labelText}</Label>
+	<select {id} bind:value={selected} on:input on:keydown={handleKeydown}>
```
peddermaster2, line 22: Does this `{id}` in both lines result in having the same ID on the `<label>` and the `<select>`?

If so, this is wrong. You can not have duplicate IDs in HTML.

PurpurTentakel97: No inside the Label it turns into the `for` argument.
So no duplicate IDs here.
But would you rename it to `for` that and just write `for={id}`.
The input will have the same problem.

src/elements/navigation/footer.svelte
PurpurTentakel97: thats right.

src/elements/input/dropDown.svelte
peddermaster2: Yeah. Otherwise it is confusing. And the label can have an ID.

mgerhold (approved):

Issue #63, opened 2024-10-19
include calendar download link


mgerhold:
On the old website, we had this link:
![image](https://github.com/user-attachments/assets/f637d810-39bf-4684-8529-0c1c582d946f)

The new website should also have this. But we have to discuss where we put the .ics file that should be downloaded. We could even auto-generate it in the backend each time someone wants to download it.

This doesn't have to be implemented immediately, but I create this issue so that we don't forget it.

Issue #43, opened 2024-09-23
popup overflow


PurpurTentakel97:
popup overflows when zommed in.

PR #75, opened 2024-10-28
CSS stuff

### LINKED ISSUE #36: discuss paragraph font

### LINKED ISSUE #44: fontsize in person popup

### LINKED ISSUE #69: Turn html font size to 62,5%

### LINKED ISSUE #70: use css variables for margin padding and font-size

### LINKED ISSUE #71: use class for fonts

PurpurTentakel97:
### features
- use global CSS variables to control margin, padding, gap, font-size, line-height. This should make the side more clean.
- extract fonts into global CSS classes.
- use a font-size of 62,5% 1rem == 10px if browser font-size is 16px) as default. This allows to calculate pixels verry easy uses the users font-size settings. Our old font-size did override the users settings.
- user more clear names for CSS classes

### issues
- close #71 
- close #70 
- close #69 

### maybe fixed
I need a opinion if this issues are fixed with this PR or not. 
- close #36 
- #43
- close #44 

### hint
- there will be a formatting PR. So no need to comment on formatting.
- not implemented: #63

### DIFF
--- a/src/app.html
+++ b/src/app.html
@@ -3,6 +3,7 @@
 	<head>
 		<meta charset="utf-8" />
 		<link rel="icon" href="%sveltekit.assets%/favicon.png" />
+		<link rel="stylesheet" href="%sveltekit.assets%/Spacings.css"/>
 		<link rel="stylesheet" href="%sveltekit.assets%/Style.css"/>
 		<link rel="stylesheet" href="%sveltekit.assets%/Hover.css"/>
 		<link rel="stylesheet" href="%sveltekit.assets%/Fonts.css"/>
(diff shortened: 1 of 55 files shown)


mgerhold (commented):
For me, the paragraph font is now okay (I think). Also the font in the popup is okay.
Has the size of the headings increased? I think those are just too big in comparison to the paragraphs.

I noticed that the font in the header broke:
image

Regarding #43: When I zoom in more than 120 %, the popups still overflow. In theory, nobody would try to zoom in that much because the box gets really huge. But on the other hand, the box is also almost empty:
image
There's just so much free and unused space. Maybe the boxes should be smaller? Or we should increase the font even more (not to increase readability, but to "fill more space"). What do you think?

mgerhold (approved):

PR #78, opened 2024-11-01
Dev

### LINKED ISSUE #63: include calendar download link

### LINKED ISSUE #72: tabs wrong index after reload
to provoke this navigate to some other tab than the left (index 0) one and reload the page.
the tab will switch to index 1 but the reloaded page is nor corresponding anymore.

### LINKED ISSUE #76: Heading Size
Adjust the heading size.
it seems to be a litte bit big.

PurpurTentakel97:
### features
- reduces the `font-size` of the headings
- moves the `LinkWithIcon` into a separate component to be able to reuse it.
- display ics download link on main page. 
- tabs now checking if they display the correct current state based on the path.
    - to check this just navigate to a non default page and reload with F5.

### issues
- close #63 
- close #76 
- close #72

### DIFF
--- a/src/routes/year/[year]/year.svelte
+++ b/src/routes/year/[year]/year.svelte
@@ -17,8 +17,10 @@
 	import Paragraph from 'elements/text/paragraph.svelte';
 	import SponsorArray from 'elements/sponsor/sponsorArray.svelte';
 	import Schedule from 'elements/schedule/schedule.svelte';
+	import LinkWithIcon from 'elements/input/linkWithIcon.svelte';
 
 	import { formatDate } from 'helper/dates';
+	import { apiUrl } from 'helper/links';
 
 	export let data: LoadYear;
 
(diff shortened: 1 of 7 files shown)


src/routes/year/[year]/year.svelte
```diff
@@ -131,6 +135,14 @@
 
 		<Section id="Schedule">
 			<HeadlineH2 classes="headline-h2-border">Plan</HeadlineH2>
+			<LinkWithIcon
+				href={apiUrl(`/api/events/${data.year.event.year}/ics`)}
+				title="Klicke um den Ablaufplan als ICS-Datei herunter zu laden"
+				icon="Calender"
+				newTab={false}
+			>
+				Verpasse keinen Vortrag und hole dir jetzt alle Termine in deinen Kalendar. Klick hier
```
Comment (mgerhold) [inline src/routes/year/[year]/year.svelte:144]:
```suggestion
				Verpasse keinen Vortrag und hole dir jetzt alle Termine in deinen Kalender. Klicke hier!
```

mgerhold (approved):

Issue #102, opened 2024-12-16
initial cropped picure


PurpurTentakel97:
The initial cropped picture needs to be resized.
right now it is just the size the picture has. 
Im not able to do that without some weird scaling stuff is happening.

PurpurTentakel97:
implemented in #103 already

PR #103, opened 2024-12-16
Speaker team member

### LINKED ISSUE #93: implement social media mechanic

PurpurTentakel97:
close #93 

This introduces the dashboard for Team Member and Speaker.
it includes the Event data and the social media links.

disclaimer.
Im not able to scale the cropped picture in a right way after uploading before the backend cropped it. maybe that is something we can discuss on Wednesday if there is a way. 
I open this issue for that. #102


mgerhold (commented):
Besides the comments (mostly grammar stuff), I also played around with the implementation and have some additional notes and findings:
* For warning messages, should we use black font instead of white? White on yellow is hard to read. 
image
* The font in dropdowns is hard to read. Maybe we can use the same font like the current selection? 
image
* The button for the cropper should not be an "X" (because this looks like all changes are discared), but rather say "Okay" or "Speichern" or something similar. Maybe it should then also be placed unter the cropper. Or there should be an additional button so that the user can either accept the cropped image or discard the new photo alltogether.
* "Dieser Datensatz ist freigegeben" should maybe disappear as soon as the user makes a change (either textual or by uploading a new photo).
* I don't think there should be two separate warning messages for "Dieser Datensatz muss noch freigegeben werden" *and* the requested changes (because, if there are changes requested, it should be clear that the data is not approved yet). Maybe we can just hide the first warning if the second is present, but then extend the message itself (something like: "Der Datensatz wurde aus dem folgenden Grund nicht freigegeben: <requested changes>"
* The requested changes do not disappear after submitting new data.
* Outstanding data approvals don't disappear upon approving them.
* After approving new data, the browser's address list still shows its anchor (e.g. `http://localhost:8080/dashboard/admin/approval#dashboard-admin-approval-1-4`). Pressing F5 then leads to a non-existing anchor. I guess the whole page should be reloaded (?).
* The images (and other data) on the frontpage are not reloaded when clicking the Test-Conf logo (upper left corner). One has to press F5 for the changes to show.
* When creating a new social media link, but clicking "Löschen" before saving it, an error occurs.
* The requested changes for social media links need an explanatory text, e.g. "Der folgende Social-Media-Link wurde abgelehnt: <requested changes>"
* There's no information about whether the social media links already have been approved.
* When creating a new social media link and then leaving the page without saving, there is no popup that prevents this.

Two more general comments (not related to this pull request):
* Not directly related to this pull request, but can we change the order of tabs in the admin dashboard? I think that they should be ordered according to their importance (approvals first, then events, then globals).
* Before releasing the page to the public, we should add many explanatory texts. But we should do this in one go at the end.

PurpurTentakel97:
> The requested changes do not disappear after submitting new data.

This should disappear when the page is reloaded and the new data gets provided. It think this should be good because maybe you want to save halfway and than the requested changes are already gone. 

> Outstanding data approvals don't disappear upon approving them.

This is Admin Dashboard ( #100 )

> After approving new data, the browser's address list still shows its anchor (e.g. 
http://localhost:8080/dashboard/admin/approval#dashboard-admin-approval-1-4). Pressing F5 then leads to a non-existing anchor. I guess the whole page should be reloaded (?).

Admin Dashboard ( #104 )

> The images (and other data) on the frontpage are not reloaded when clicking the Test-Conf logo (upper left corner). One has to press F5 for the changes to show.

Is that a big issue?
That effects only us. The browser will cash these things I think. 

> There's no information about whether the social media links already have been approved.

How should this look like? Every link has its own approval state.

> When creating a new social media link and then leaving the page without saving, there is no popup that prevents this.

Because you did not enter any data. the bare link does nothing thats right. 

> Not directly related to this pull request, but can we change the order of tabs in the admin dashboard? I think that they should be ordered according to their importance (approvals first, then events, then globals).

#105

> Before releasing the page to the public, we should add many explanatory texts. But we should do this in one go at the end.

#106

mgerhold:
> > The requested changes do not disappear after submitting new data.
> 
> This should disappear when the page is reloaded and the new data gets provided. It think this should be good because maybe you want to save halfway and than the requested changes are already gone.

I wouldn't fully agree here since I don't think users will save halfway. But I don't think it's a real issue.



> > The images (and other data) on the frontpage are not reloaded when clicking the Test-Conf logo (upper left corner). One has to press F5 for the changes to show.
>
> Is that a big issue?
> That effects only us. The browser will cash these things I think.

Why should this only affect us? If the user tries to refresh the page by clicking the logo (instead of using F5), maybe because they're waiting for a new speaker announcement, they will not see any change.
I think this is not caused by the caching of the browser. I assume that Svelte doesn't reload anything because it knows that you're trying to show the page that you're already on. But I'm not sure. I'm also not sure how important this is. Maybe not a real issue? Not sure.

> > There's no information about whether the social media links already have been approved.
>
> How should this look like? Every link has its own approval state.

Maybe just a little green checkmark and a text next to the link (instead of above the link), e.g. "✅ akzeptiert"?

> > When creating a new social media link and then leaving the page without saving, there is no popup that prevents this.
>
> Because you did not enter any data. the bare link does nothing thats right.

Okay, fair enough, that's okay.

PurpurTentakel97:
> Why should this only affect us? If the user tries to refresh the page by clicking the logo (instead of using F5), maybe because they're waiting for a new speaker announcement, they will not see any change.
I think this is not caused by the caching of the browser. I assume that Svelte doesn't reload anything because it knows that you're trying to show the page that you're already on. But I'm not sure. I'm also not sure how important this is. Maybe not a real issue? Not sure.

Oh now I get that.
How did you even check that? 🙈 
But yes. Maybe svelte can reload the page all together.

mgerhold:
> How did you even check that? 🙈

Well...I opened the page twice (in two different tabs) 😅

PurpurTentakel97:
> I don't think there should be two separate warning messages for "Dieser Datensatz muss noch freigegeben werden" and the requested changes (because, if there are changes requested, it should be clear that the data is not approved yet). Maybe we can just hide the first warning if the second is present, but then extend the message itself (something like: "Der Datensatz wurde aus dem folgenden Grund nicht freigegeben: "

This is no longer working because we did this toggle in the last stream.
The requested changes would dissipier when there are unsaved changes.
So I will ignore that.

mgerhold (approved):

PR #112, opened 2024-12-27
Jet brains

PurpurTentakel97:
This pr addresses nearly all of the jetbrains hints.
There is just one A11y warning I am not able to disable.
I don't see a problem there because the provided button does the same so all how needs a11y functions will be able to use the button. Thy don't miss out anything.


src/pages/speakerTeamMemberEvent.svelte
```diff
@@ -130,7 +131,7 @@
 		}
 	});
 
-	function updateCropperData(e: any): void {
+	function updateCropperData(e: typeof Cropper.prototype.$$events_def.cropcomplete): void {
```
mgerhold, line 134: `typeof Cropper.prototype.$$events_def.cropcomplete` 😂
TypeScript is beautiful 😂

mgerhold (approved):

PR #114, opened 2024-12-28
109 check https prefix on social media links

### LINKED ISSUE #109: check https prefix on social media links
add https:// prefix if not provided.
check for https:// and http://

PurpurTentakel97:
close #109 

This checks if a provided social media URL starts with `https://` or `http://`.
If not it add `https://` in front of the link an checks it.
I decided to manipulate the original data to make sure one can see that the validation is done with the motived data. One can change the URL properly when something went wrong. In case something like `https://htps://www.google.de` happend. One could see that imminently than. 

Formatter did his thing.
I did 2 commits. Maybe check only the first one or only have a look on the `verify()` function.
There are no other changes here.

### DIFF
--- a/src/pages/speakerTeamMemberSocialMediaLinks.svelte
+++ b/src/pages/speakerTeamMemberSocialMediaLinks.svelte
@@ -1,175 +1,177 @@
 <script lang="ts">
-	import * as Menu from 'menu/dashboard';
-	import * as MenuItem from 'menu/menuItems';
+    import * as Menu from 'menu/dashboard';
+    import * as MenuItem from 'menu/menuItems';
 
-	import type { LoadDashboard, LoadUserSocials } from 'types/dashboardLoadTypes';
-	import type { DashboardSocialMediaLink } from 'types/dashboardProvideTypes';
-	import type {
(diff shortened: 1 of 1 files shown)


src/pages/speakerTeamMemberSocialMediaLinks.svelte
```diff
@@ -43,19 +43,28 @@
 	}
 
 	function validate(): string[] {
+		const https:string = 'https://';
+		const http:string = 'http://';
```
mgerhold: Is this formatting "correct"? It seems to contradict the usual style.
I'd expect `const https: string` instead of `const https:string`.

PurpurTentakel97: Did you look at the first commit?
Please remember that the formatter of cores did not run yet.
I only run the formatter on the second commit to split the changes from the formatter.

So on the second commit the formatting should be just fine.

mgerhold: Sorry, my fault!

mgerhold (approved):

PR #115, opened 2024-12-29
add facebook icon

PurpurTentakel97:
add the facebook icon.

I decided to use the round facebook icon instead of the raw f because the raw f is not squared.
That looked kinda weird.
Now Facebook ist the only round icon but I feel it fits better.

If you want to check the raw f for yourself:
- check out this branch
- replace `fa-facebook` with `fa-facebook-f` in the icon lookup
- add a facebook type somehow. Easiest was would be to seed this I think.


mgerhold (approved):
I compared the two icon variants. I also prefer the one that you chose 👍

PR #116, opened 2025-01-01
User

PurpurTentakel97:
this introduces a call-for-papaers page.


src/lookup/propagationLookup.ts
```diff
@@ -0,0 +1,19 @@
+type PropagationDict = {
```
mgerhold, line 1:
```suggestion
type ApplicationDict = {
```

src/lookup/propagationLookup.ts
```diff
@@ -0,0 +1,19 @@
...
+    [key: string]: string;
+};
+
+
+const lookup: PropagationDict = {
```
mgerhold, line 6: Should we change this somehow? It feels a bit weird to use the whole error message of the backend here. Should we use simpler strings then? I doesn't really make a big difference, but it feels unclean this way. What do you think? I'm also fine with leaving it like that.

src/lookup/propagationLookup.ts
```diff
@@ -0,0 +1,19 @@
...
+    'User cannot apply as speaker since they already are a speaker.':              'Du hast dich bereits beworben.\nDaher kannst du dich nicht erneut bewerben.\nWenn du bereits ein Speaker bist, kannst du deine Daten im Speaker Tab bearbeiten.',
```
mgerhold, line 7:
```suggestion
    'User cannot apply as speaker since they already are a speaker.':              'Du hast dich bereits beworben.\nDaher kannst du dich nicht erneut bewerben.\nWenn du bereits ein Speaker bist, kannst du deine Daten im Speaker-Tab bearbeiten.',
```

src/lookup/propagationLookup.ts
```diff
@@ -0,0 +1,19 @@
...
+    'Speaker applications are not accepted at this time.':                         'Aktuell kann man sich nicht bewerben.\nSchau gerne bei der nächsten Bewerbungsphase wieder vorbei.\nWenn du bereits ein Speaker bist, kannst du deine Daten im Speaker Tab bearbeiten.',
```
mgerhold, line 8:
```suggestion
    'Speaker applications are not accepted at this time.':                         'Aktuell kann man sich nicht bewerben.\nSchau gerne bei der nächsten Bewerbungsphase wieder vorbei.\nWenn du bereits ein Speaker bist, kannst du deine Daten im Speaker-Tab bearbeiten.',
```

src/lookup/propagationLookup.ts
```diff
@@ -0,0 +1,19 @@
...
+    'User cannot apply as speaker since they already have a pending application.': 'Du hast dich bereits für dieses Jahr beworben\nWir sind gerade dabei deine Bewerbung zu überprüfen.\nSobald wir deine Bewerbung angenommen haben, kannst du deinen Eintrag im Speaker Dashboard bearbeiten.',
```
mgerhold, line 9:
```suggestion
    'User cannot apply as speaker since they already have a pending application.': 'Du hast dich bereits für dieses Jahr beworben\nWir sind gerade dabei, deine Bewerbung zu überprüfen.\nSobald wir deine Bewerbung angenommen haben, kannst du deinen Eintrag im Speaker-Dashboard bearbeiten.',
```
Besides the suggestion above: Why is it "Dashboard" here, but "Tab" in the other messages?

src/lookup/propagationLookup.ts
```diff
@@ -0,0 +1,19 @@
...
+};
+
+
+export function propagationLookup(key: string): string {
```
mgerhold, line 13:
```suggestion
export function applicationLookup(key: string): string {
```

src/pages/speakerTeamMemberEventForm.svelte
```diff
@@ -0,0 +1,222 @@
...
+    import CroppedImage from 'elements/image/croppedImage.svelte';
+    import Button from 'elements/input/button.svelte';
+    import Image from 'elements/image/image.svelte';
+    import Input from 'elements/input/input.svelte';
+    import TextArea from 'elements/input/textArea.svelte';
+    import MyCropper from 'elements/MyCropper/myCropper.svelte';
+    import FormWrapper from 'elements/wrapper/formWrapper.svelte';
+
+    export let data: LoadSpeakerTeamMemberEvent;
+    // If the save button is not displayed, the form also gets removed. This the on:save will not be triggert is that case.
```
mgerhold, line 20: > This the on:save will not be triggert is that case.

This is not a valid sentence 😅 What do you mean by this?

src/pages/speakerTeamMemberSocialMediaLinkForm.svelte
```diff
@@ -0,0 +1,90 @@
...
+    import { setUnsavedChanges } from 'stores/saved';
+    import { getIDFromSocialMediaType } from 'pageHelper/speakerTeamMemberSocials';
+    import { createEventDispatcher } from 'svelte';
+
+    import EditSocialMedia from 'elements/input/editSocialMedia.svelte';
+    import Button from 'elements/input/button.svelte';
+    import GeneralPopup from 'elements/popups/generalPopup.svelte';
+    import FormWrapper from 'elements/wrapper/formWrapper.svelte';
+
+    // If the save button is not displayed, the form also gets removed. This the on:save will not be triggert is that case.
```
mgerhold, line 14: Same here

src/routes/dashboard/user/application/+page.svelte
```diff
@@ -0,0 +1,253 @@
...
+    export let data: LoadDashboard & LoadUserApplication;
+
+    let messages: string[] = [];
+    let success: boolean   = false;
+    let eventForm: SpeakerTeamMemberEventForm;
+    let saveMessage: SaveMessage;
+
+    function deleteLink(e: CustomEvent<number>) {
+        if (!data.data) {
+            console.error('data ist undefined');
```
mgerhold, line 38:
```suggestion
            console.error('data is undefined');
```

src/routes/dashboard/user/application/+page.svelte
```diff
@@ -0,0 +1,253 @@
...
+      classes="navigation-tabs-dashboard-subpage"
+/>
+<UnsavedChangesCallbackWrapper callback={trySave} />
+
+<SectionDashboard classes="standard-dashboard-section">
+    {#if success}
+        <div class="dashboard-user-application-error">
+            <Paragraph classes="paragraph-pre-wrap"
+                       --text-align="center">
+                {"Du hast dich erfolgreich beworben.\nWir werden nun deine Bewerbung kontrollieren.\nWenn wir dich annehmen kannst du deine Daten im Speaker Dashboard bearbeiten.\nDazu wird dann ein neuer Tab oben rechts erscheinen."}
```
mgerhold, line 143:
```suggestion
                {"Du hast dich erfolgreich beworben.\nWir werden nun deine Bewerbung kontrollieren.\nWenn wir dich annehmen, kannst du deine Daten im Speaker-Dashboard bearbeiten.\nDazu wird dann ein neuer Tab oben rechts erscheinen."}
```

src/routes/dashboard/user/application/+page.svelte
```diff
@@ -0,0 +1,253 @@
...
+                {/each}
+            </div>
+            <div class="dashboard-user-application-section">
+                <HeadlineH2>Deine Bewerbung</HeadlineH2>
+                <SpeakerTeamMemberEventForm bind:this={eventForm}
+                                            bind:data={data.data.speaker}
+                                            displaySaveButton={false} />
+            </div>
+            <div class="dashboard-user-application-section">
+                <HeadlineH2>Social Media-Daten</HeadlineH2>
```
mgerhold, line 199:
```suggestion
                <HeadlineH2>Social-Media-Daten</HeadlineH2>
```
Or we just say "Social Media" (then we don't need the dash).

src/types/dashboardProvideTypes.ts
```diff
@@ -143,3 +143,14 @@ export type DashboardSocialMediaLinkType = z.infer<typeof dashboardSocialMediaLi
 
 export const dashboardAllSocialMediaLinkTypeScheme = z.array(dashboardSocialMediaLinkTypeScheme);
 export type DashboardAllSocialMediaLinkType = z.infer<typeof dashboardAllSocialMediaLinkTypeScheme>;
+
+export const dashboardEventPropagationScheme = z.object({
```
mgerhold, line 147:
```suggestion
export const dashboardEventApplicationScheme = z.object({
```

src/types/dashboardProvideTypes.ts
```diff
@@ -143,3 +143,14 @@ export type DashboardSocialMediaLinkType = z.infer<typeof dashboardSocialMediaLi
...
+                                                            id:                    z.number(),
+                                                            title:                 z.string(),
+                                                            subtitle:              z.string(),
+                                                            start_date:            z.string(),
+                                                            end_date:              z.string(),
+                                                            call_for_papers_start: z.string(),
+                                                            call_for_papers_end:   z.string(),
+                                                        });
+export type DashboardEventPropagation = z.infer<typeof dashboardEventPropagationScheme>;
```
mgerhold, line 156:
```suggestion
export type DashboardEventApplication = z.infer<typeof dashboardEventPropagationScheme>;
```

src/lookup/propagationLookup.ts
PurpurTentakel97: It always feels unclean. (in registry too for example)
I suggested a Code system once.
But you did not like it.

I would be perfectly fine when you would provide something like.

```
{
    "error": "Hier goes the message",
    "code": 301,
}
```

I could suggest a code test-conf-error-code we can use.
Just like 3XX for propagation.
301 - No current event
302 - already applied
etc....

So I would just define a tabel just like a http code.
But only for errors or would you like to have success codes two?

What do you think?

src/pages/speakerTeamMemberEventForm.svelte
PurpurTentakel97: I mean by this, that I was verry tired. LUL 

No I mean that when the button is not displayed, the form does get removed. 

This allows to use the input tags in a different context.
But you have to provide a form and a button again in order to be able to trigger the form again.

mgerhold: Okay, I don't think that I'm deep enough into this to fully understand it, but that's fine. Then maybe just fix the comment so that your future self will understand what's going on 😅

src/lookup/propagationLookup.ts
mgerhold: I would not give you both an error message and a number, since that would be redundant. The error number alone would be bad, too, because then you'd have to find out the meaning of each number.
I'm thinking more of something that looks like an enum, e.g. `NO_CURRENT_EVENT`, `ALREADY_APPLIED` etc.
But then, on the other hand, I wonder if that really is a benefit over the current solution. I'm unsure here.

PurpurTentakel97: I don't know.

I mean the only benefits that the shorter "Enum-code" has is that the risk of typos is not that high.
But for my part I copy paste you're message. 
So Im fine with both.

mgerhold: I think there's another benefit: When I use some kind of "enum code", then it's obvious (when looking at the backend code) that those string constants must not be changed because they are used by the frontend. With a "human readable" error message, this is not really clear. And then, when one changes the error messages in the backend, this breaks the frontend logic.
I implemented this here: https://github.com/TechStreamConference/website-backend/pull/100
Please look at it carefully to not miss some spots in the frontend that have to be changed now.

mgerhold (approved):

mgerhold (approved):

PR #119, opened 2025-01-04
add fixed width for person rectangle

### LINKED ISSUE #118: wrong image size on front page
We have bug across both the frontend and the backend: On the [original website](https://www.test-conf.de), all photos are 350x350 pixels in size. However, on our site all photos are 300x300. So I think there's two things we have to do:
- the frontend has to display the images at the correct size *regardless of their actual resolution*
- the backend has to fix the target resolution (I will track this in [this issue](https://github.com/TechStreamConference/website-backend/issues/96))

Currently, the bug can lead to layout problems:
image

PurpurTentakel97:
close #118 

I use a 100% width on the person rectangle now.
It seems to work.
Please verify that.

### DIFF
--- a/src/elements/person/personRectangle.svelte
+++ b/src/elements/person/personRectangle.svelte
@@ -49,6 +49,7 @@
         border-radius:    var(--border-radius);
         transition:       var(--transition-duration);
         text-align:       center;
+        width:            100%;
     }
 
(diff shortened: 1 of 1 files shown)


mgerhold (approved):
Looks good 👍
image

PR #120, opened 2025-01-06
scroll to top after any safe action

### LINKED ISSUE #117: feature request: scroll up after saving data
When editing a speaker entry and then clicking the save button, at the top of the page a message appears that confirms that the data has been saved successfully. But the user probably will not see that message, because they scrolled down to the save button (and the window is not big enough to fit the whole page).
See this screenshot for example:
image
When I click save, I cannot see the success message (this is on 100 % zoom level).

PurpurTentakel97:
close #117 

the event page is now scrolling to top on any safe action


mgerhold (approved):
Nice, thanks! 👍

PR #124, opened 2025-01-08
Reset password

PurpurTentakel97:
this introduces the reset password mechanic

btw. is there a way to stick to one key for errors?
I think you stick to one key but code igniter switches.
here e.g. it uses `new_password` because its the field with the error I guess.

So I try to get all messages. This ends in this
```
            const entriesAsync = async (response: Response): Promise<string[]> => {
                const text: string     = await response.text();
                const json: {
                    [key: string]: string
                }                      = JSON.parse(text);
                const values: string[] = Object.values(json);
                let toReturn: string[] = [];
                for (const value of values) {
                    toReturn.push(registerLookup(value));
                    console.error('error while register from dashboard: ' + value);
                }
                return toReturn;
            };
``` 

Or to you have a better Idea because this looks kinda wonky.

### DIFF
--- a/src/lookup/registerLookup.ts
+++ b/src/lookup/registerLookup.ts
@@ -25,7 +25,7 @@ const lookup: RegisterDict = {
     PASSWORD_MISSING_SPECIAL_CHARACTER: 'Deinem Passwort fehlt ein Sonderzeichen.',
 
     // token
-    TOKEN_EXPIRED_OR_DOES_NOT_EXIST : 'Der URL-Token existiert nicht oder ist abgelaufen.'
+    TOKEN_EXPIRED_OR_DOES_NOT_EXIST: 'Der Link ist ungültig oder bereits abgelaufen.',
 };
 
 export function registerLookup(key: string): string {
--- a/src/lookup/registerLookup.ts
+++ b/src/lookup/registerLookup.ts
@@ -22,7 +22,10 @@ const lookup: RegisterDict = {
     PASSWORD_MISSING_LOWERCASE_LETTER:  'Deinem Passwort fehlt ein Kleinbuchstabe.',
     PASSWORD_MISSING_UPPERCASE_LETTER:  'Deinem Passwort fehlt ein Großbuchstabe.',
     PASSWORD_MISSING_NUMBER:            'Deinem Passwort fehlt eine Ziffer.',
-    PASSWORD_MISSING_SPECIAL_CHARACTER: 'Deinem Passwort fehlt ein Sonderzeichen.'
+    PASSWORD_MISSING_SPECIAL_CHARACTER: 'Deinem Passwort fehlt ein Sonderzeichen.',
+
+    // token
+    TOKEN_EXPIRED_OR_DOES_NOT_EXIST : 'Der URL-Token existiert nicht oder ist abgelaufen.'
 };
 
 export function registerLookup(key: string): string {
--- a/src/routes/login/+page.svelte
+++ b/src/routes/login/+page.svelte
@@ -71,7 +71,7 @@
         };
 
         if (data.username_or_email.length === 0) {
-            errorMessage = 'Keine E-Mail oder Namen angegeben.';
+            errorMessage = 'Keine E-Mail-Adresse oder Name angegeben.';
             return;
         }
 
@@ -84,8 +84,6 @@
             changeState(State.ResetPasswordSuccess);
             return;
         }
-
-        console.log(response);
 
         try {
             const scheme = z.object({
@@ -140,7 +138,7 @@
                 Noch keinen Account?
             </Link>
             <Link href="#"
-                  title="Klicke, um dein Passwort zurück zu setzen."
+                  title="Klicke, um dein Passwort zurückzusetzen."
                   on:click={() => {changeState(State.ResetPasswordPending)}}
             >
                 Passwort vergessen?
@@ -161,7 +159,7 @@
                   type="text"
                   labelText="Nutzername oder E-Mail:"
                   placeholderText="Nutzername oder E-Mail"
-                  ariaLabel="Gib den Nutzernamen oder die E-Mail ein"
+                  ariaLabel="Gib den Nutzernamen oder die E-Mail-Adresse ein"
                   bind:value={usernameOrEmail}
             />
             <Button classes="login-button"
@@ -169,7 +167,7 @@
                     ariaLabel="Klicke, um dein Passwort zurückzusetzen">Passwort zurücksetzen
             </Button>
             <Link href="#"
-                  title="Klicke, wieder zum login zu kommen."
+                  title="Klicke, um wieder zur Anmeldung zu kommen."
                   on:click={() => {changeState(State.Login)}}
             >
                 zurück
@@ -180,7 +178,7 @@
               on:submit|preventDefault={() => {changeState(State.Login);}}>
             <HeadlinePage>Passwort zurückgesetzt</HeadlinePage>
             <Paragraph classes="paragraph-pre-wrap"
-                       --text-align="center">{'Dein Passwort wurde erfolgreich zurück gesetzt.\nDu erhältst eine Mail mit einem Link.\nDort kannst du dein neues Passwort eingeben.\nDanach kannst du dich mit deinem neuen Passwort wieder anmelden.'}</Paragraph>
+                       --text-align="center">{'Wir haben dir eine E-Mail gesendet. Darin findest du einen Link, mit dessen Hilfe du ein neues Passwort auswählen kannst.\nDanach kannst du dich mit deinem neuen Passwort wieder anmelden.'}</Paragraph>
             <Button classes="login-button"
                     type="submit"
                     ariaLabel="Klicke, um dich anzumelden">Login
--- a/src/routes/login/+page.svelte
+++ b/src/routes/login/+page.svelte
@@ -9,6 +9,7 @@
     import Input from 'elements/input/input.svelte';
     import Button from 'elements/input/button.svelte';
     import Link from 'elements/text/link.svelte';
+    import Paragraph from 'elements/text/paragraph.svelte';
 
     import { goto } from '$app/navigation';
     import { apiUrl } from 'helper/links';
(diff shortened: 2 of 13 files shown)


mgerhold:
I think it would be possible to wrap CodeIgniter's errors inside the `error` key. For example, currently the `register` endpoint may return something like this:
```json
{
  "password": "PASSWORD_TOO_SHORT"
}
```
I could wrap it so that it looks like this:
```json
{
    "error": {
        "password": "PASSWORD_TOO_SHORT"
    }
}
```
But that would mean that the value won't necessarily always be a string, but sometimes an array instead. I don't think we really can avoid the array, since CodeIgniter might give multiple errors at once (e.g. password too short, username missing, etc.)
So you'd have to use a loop anyway, wouldn't you?

Another possibility would be that the backend joins all the errors together into one single string. I don't really know if I would like that, though 🙈

PurpurTentakel97:
No I dont think this would help ether.
Maybe I should extract this llambda and than just call it.
Seems to be the best way. What do you think?

PurpurTentakel97:
But that would mean I should maybe check if I used that everywhere or if I use a Zod Scheme at some point.

mgerhold:
I don't really know what the best solution would be on your side :/
Try to avoid code duplication, sure, but if it's not possible, then that just how it is 🤷‍♂️

src/lookup/registerLookup.ts
```diff
@@ -22,7 +22,10 @@ const lookup: RegisterDict = {
     PASSWORD_MISSING_LOWERCASE_LETTER:  'Deinem Passwort fehlt ein Kleinbuchstabe.',
     PASSWORD_MISSING_UPPERCASE_LETTER:  'Deinem Passwort fehlt ein Großbuchstabe.',
     PASSWORD_MISSING_NUMBER:            'Deinem Passwort fehlt eine Ziffer.',
-    PASSWORD_MISSING_SPECIAL_CHARACTER: 'Deinem Passwort fehlt ein Sonderzeichen.'
+    PASSWORD_MISSING_SPECIAL_CHARACTER: 'Deinem Passwort fehlt ein Sonderzeichen.',
+
+    // token
+    TOKEN_EXPIRED_OR_DOES_NOT_EXIST : 'Der URL-Token existiert nicht oder ist abgelaufen.'
```
mgerhold, line 28:
```suggestion
    TOKEN_EXPIRED_OR_DOES_NOT_EXIST : 'Der Link ist ungültig oder bereits abgelaufen.'
```

src/routes/login/+page.svelte
```diff
@@ -48,12 +64,51 @@
...
         errorMessage       = await entriesAsync(response);
     }
+
+    async function resetPassword(): Promise<void> {
+        const data = {
+            username_or_email: usernameOrEmail.trim(),
+        };
+
+        if (data.username_or_email.length === 0) {
+            errorMessage = 'Keine E-Mail oder Namen angegeben.';
```
mgerhold, line 74:
```suggestion
            errorMessage = 'Keine E-Mail-Adresse oder Name angegeben.';
```

src/routes/login/+page.svelte
```diff
@@ -48,12 +64,51 @@
...
+            body:   JSON.stringify(data),
+        });
+
+        if (response.ok) {
+            changeState(State.ResetPasswordSuccess);
+            return;
+        }
+
+        console.log(response);
+
```
mgerhold, line 89: I would not log this. We can use the network tab in the developer tools to see the response.
```suggestion

```

src/routes/login/+page.svelte
```diff
@@ -84,31 +139,57 @@
                   title="Klicke, um einen neuen Account anzulegen">
                 Noch keinen Account?
             </Link>
+            <Link href="#"
+                  title="Klicke, um dein Passwort zurück zu setzen."
```
mgerhold, line 143:
```suggestion
                  title="Klicke, um dein Passwort zurückzusetzen."
```

src/routes/login/+page.svelte
```diff
@@ -84,31 +139,57 @@
...
+              on:submit|preventDefault={resetPassword}>
+            <HeadlinePage>Passwort zurücksetzen</HeadlinePage>
+            <ErrorMessage message={errorMessage} />
+            <Input
+                  classes="login-username-mail login-username-mail-reset-extra"
+                  id="login-username-or-email"
+                  type="text"
+                  labelText="Nutzername oder E-Mail:"
+                  placeholderText="Nutzername oder E-Mail"
+                  ariaLabel="Gib den Nutzernamen oder die E-Mail ein"
```
mgerhold, line 164:
```suggestion
                  ariaLabel="Gib den Nutzernamen oder die E-Mail-Adresse ein"
```

src/routes/login/+page.svelte
```diff
@@ -84,31 +139,57 @@
...
+                  bind:value={usernameOrEmail}
+            />
+            <Button classes="login-button"
+                    type="submit"
+                    ariaLabel="Klicke, um dein Passwort zurückzusetzen">Passwort zurücksetzen
+            </Button>
+            <Link href="#"
+                  title="Klicke, wieder zum login zu kommen."
```
mgerhold, line 172:
```suggestion
                  title="Klicke, um wieder zur Anmeldung zu kommen."
```

src/routes/login/+page.svelte
```diff
@@ -84,31 +139,57 @@
...
+            >
+                zurück
+            </Link>
+        </form>
+    {:else if state === State.ResetPasswordSuccess}
+        <form class="login-form-width-wrapper login-form-gap-wrapper"
+              on:submit|preventDefault={() => {changeState(State.Login);}}>
+            <HeadlinePage>Passwort zurückgesetzt</HeadlinePage>
+            <Paragraph classes="paragraph-pre-wrap"
+                       --text-align="center">{'Dein Passwort wurde erfolgreich zurück gesetzt.\nDu erhältst eine Mail mit einem Link.\nDort kannst du dein neues Passwort eingeben.\nDanach kannst du dich mit deinem neuen Passwort wieder anmelden.'}</Paragraph>
```
mgerhold, line 183:
```suggestion
                       --text-align="center">{'Wir haben dir eine E-Mail gesendet. Darin findest du einen Link, mit dessen Hilfe du ein neues Passwort auswählen kannst.\nDanach kannst du dich mit deinem neuen Passwort wieder anmelden.'}</Paragraph>
```

src/routes/login/+page.svelte
```diff
@@ -84,31 +139,57 @@
...
+            <Button classes="login-button"
+                    type="submit"
+                    ariaLabel="Klicke, um dich anzumelden">Login
```
mgerhold, line 186: On the other buttons, the text is "Anmelden" instead of "Login".
```suggestion
                    ariaLabel="Klicke, um dich anzumelden">Anmelden
```

src/routes/reset-password/+page.svelte
```diff
@@ -0,0 +1,137 @@
...
+    import PageWrapper from 'elements/section/pageWrapper.svelte';
+    import PasswordHints from 'elements/text/passwordHints.svelte';
+    import HeadlinePage from 'elements/text/headlinePage.svelte';
+    import Paragraph from 'elements/text/paragraph.svelte';
+    import StyledLink from 'elements/input/styledLink.svelte';
+    import Message from 'elements/text/message.svelte';
+    import { registerLookup } from 'lookup/registerLookup.js';
+
+    enum State {
+        Resettet,
```
mgerhold, line 20: The past form of "reset" is also "reset". So you can either use `Reset` or `DidReset`.
```suggestion
        DidReset,
```

src/routes/reset-password/+page.svelte
```diff
@@ -0,0 +1,137 @@
...
+                  type="password"
+                  labelText="Passwort wiederholen:"
+                  placeholderText="Passwort wiederholen"
+                  bind:value={password2}
+                  ariaLabel="Wiederhole das neue Passwort"
+            />
+            <Button
+                  classes="reset-password-submit-button"
+                  type={'submit'}
+                  ariaLabel="Klicke hier, um dein Passwort zurück zu setzen"
```
mgerhold, line 103:
```suggestion
                  ariaLabel="Klicke hier, um dein Passwort zurückzusetzen"
```

src/routes/reset-password/+page.svelte
```diff
@@ -0,0 +1,137 @@
...
+            >
+                Passwort zurücksetzen
+            </Button>
+            <PasswordHints />
+        </form>
+    {:else if state === State.Resettet}
+        <div class="reset-password-wrapper">
+            <HeadlinePage>Passwort zurückgesetzt</HeadlinePage>
+            <Paragraph classes="paragraph-pre-wrap"
+                       --text-align="center">{"Dein Passwort wurde erfolgreich zurück gesetzt.\nDu kannst dich nun mit deinem neuen Passwort anmelden."}</Paragraph>
```
mgerhold, line 113:
```suggestion
                       --text-align="center">{"Dein Passwort wurde erfolgreich geändert.\nDu kannst dich nun mit deinem neuen Passwort anmelden."}</Paragraph>
```

src/routes/reset-password/+page.svelte
```diff
@@ -0,0 +1,137 @@
...
+                  type={'submit'}
+                  ariaLabel="Klicke hier, um dein Passwort zurück zu setzen"
+            >
+                Passwort zurücksetzen
+            </Button>
+            <PasswordHints />
+        </form>
+    {:else if state === State.Resettet}
+        <div class="reset-password-wrapper">
+            <HeadlinePage>Passwort zurückgesetzt</HeadlinePage>
```
mgerhold, line 111:
```suggestion
            <HeadlinePage>Passwort geändert</HeadlinePage>
```

src/routes/reset-password/+page.svelte
mgerhold: You marked this conversation as resolved, but didn't change anything. Was this on purpose?

mgerhold (approved):

Issue #126, opened 2025-01-10
Navigation drop down 


PurpurTentakel97:
Use navigation drop down on the rest of the page.

PurpurTentakel97:
implemented in #128

PR #127, opened 2025-01-10
Time slots

PurpurTentakel97:
This adds a page to edit time slots.
I want to use the new navigation drop down on the rest of the page two.
An issue for that is here #126 

Im not quite sure if one would check the `is_special` toggle.
But if you understand it would be good for this page.
Maybe I should add some headline for this table?


src/elements/navigation/NavigationDropDown.svelte
mgerhold, line 1: Why is this file name in PascalCase, where all the other .svelte files use camelCase?

src/elements/input/toggle.svelte
```diff
@@ -0,0 +1,26 @@
...
+    }
+</script>
+
+<button class="toggle-button"
+        type="button"
+        aria-label="{ariaLabel}"
+        on:click={onToggle}
+>
+    <Icon type={toggle ? "CircleCheck" : "CircleCross"}
+          color={toggle ? "green" : "red" } />
```
mgerhold, line 18: Maybe you can make the icons customizable from the outside?

For the toggle for the time slots, we could use these icons:
- `fa-tower-broadcast` (link [here](https://fontawesome.com/icons/tower-broadcast?f=classic&s=solid)) for live talks:
  image
- `fa-video` (link [here](https://fontawesome.com/icons/video?f=classic&s=solid)) for recordings:
  image

Would it also be possible to add a tooltip? Maybe saying something like "Live-Talk oder Aufzeichnung?"

src/routes/dashboard/admin/time-slots/+page.svelte
```diff
@@ -0,0 +1,232 @@
...
+
+        console.log(`not able to look up Event with ID ${id}`);
+        throw error(404);
+    }
+
+    function addSlot() {
+        const newSlotsData: DashboardTimeSlot = {
+            event_id:   data.currentEventID,
+            id:         0,
+            start_time: getEventFromID(data.currentEventID).start_date + 'T00:00:00', // the start of the event is used here, as it should not be too far away from the actual date.
```
mgerhold, line 57: If you make it even cooler, you could use the start date of the previous time slot and add its duration to it. But this is just a "nice to have"-feature, we don't *really* need this.

src/routes/dashboard/admin/time-slots/+page.svelte
```diff
@@ -0,0 +1,232 @@
...
+        data.currentSlots = [
+            ...data.currentSlots,
+            newSlotsData,
+        ];
+    }
+
+    async function updateDisplayed(value: string) {
+        data.currentEventID = getIDFromTitle(value);
+        data.currentSlots   = await getTimeSlotsAsync(fetch, data.currentEventID);
+    }
```
mgerhold, line 71: I think you should also hide the error message here. Otherwise it will stay visible even after changing to a different event.

src/routes/dashboard/admin/time-slots/+page.svelte
```diff
@@ -0,0 +1,232 @@
...
+
+    function moveUp(index: number) {
+        errorMessage = undefined;
+        if (index - 1 == -1) {
+            errorMessage = 'Es gibt keinen Slot vor diesem Slot.';
+            return;
+        }
+
+        if (index < 0 || index >= data.currentSlots.length) {
+            console.error(`index out of bounce | Index: ${index}`);
```
mgerhold, line 104:
```suggestion
            console.error(`index out of bounds | Index: ${index}`);
```

src/routes/dashboard/admin/time-slots/+page.svelte
```diff
@@ -0,0 +1,232 @@
...
+
+    function moveDown(index: number) {
+        errorMessage = undefined;
+        if (index + 1 == data.currentSlots.length) {
+            errorMessage = 'Es gibt kein Slot hinter diesem Slot.';
+            return;
+        }
+
+        if (index < 0 || index >= data.currentSlots.length) {
+            console.error(`index out of bounce | Index: ${index}`);
```
mgerhold, line 125:
```suggestion
            console.error(`index out of bounds | Index: ${index}`);
```

src/routes/dashboard/admin/time-slots/+page.svelte
```diff
@@ -0,0 +1,232 @@
...
+            data.currentSlots[index + 1],
+        ] = [
+            data.currentSlots[index + 1],
+            data.currentSlots[index],
+        ];
+    }
+
+    function deleteEntry(index: number) {
+        if (index < 0 || index >= data.currentSlots.length) {
+            console.error(`index out of bounce | Index: ${index}`);
```
mgerhold, line 140:
```suggestion
            console.error(`index out of bounds | Index: ${index}`);
```

src/routes/dashboard/admin/time-slots/+page.svelte
```diff
@@ -0,0 +1,232 @@
...
+                      on:input={setUnsavedChanges}
+                />
+                <Input
+                      id="dashboard-admin-time-slot-duration-{index}"
+                      labelText="Dauer:"
+                      type="number"
+                      ariaLabel="Gib die Dauer des ausgewählten Slots ein."
+                      bind:value={entry.duration}
+                      on:input={setUnsavedChanges}
+                />
```
mgerhold, line 182: I think this should be a dropdown. You can use the `GET api/talk-duration-choices` endpoint to fetch akk possible values.

src/routes/dashboard/admin/time-slots/+page.svelte
```diff
@@ -0,0 +1,232 @@
...
+                      id="dashboard-admin-time-slot-start-date-{index}"
+                      labelText="Start:"
+                      type="datetime-local"
+                      ariaLabel="Gib den Start des ausgewählten Slots ein."
+                      bind:value={entry.start_time}
+                      on:input={setUnsavedChanges}
+                />
+                <Input
+                      id="dashboard-admin-time-slot-duration-{index}"
+                      labelText="Dauer:"
```
mgerhold, line 177:
```suggestion
                      labelText="Dauer (Minuten):"
```

src/routes/dashboard/admin/time-slots/+page.svelte
```diff
@@ -0,0 +1,232 @@
...
+                      type="number"
+                      ariaLabel="Gib die Dauer des ausgewählten Slots ein."
+                      bind:value={entry.duration}
+                      on:input={setUnsavedChanges}
+                />
+                <Toggle ariaLabel="Klicke, um den special-Status des Slots zu ändern"
```
mgerhold, line 183:
```suggestion
                <Toggle ariaLabel="Klicke, um zwischen Live-Talk und Aufzeichnung zu wechseln."
```

src/routes/dashboard/admin/time-slots/+page.svelte
```diff
@@ -0,0 +1,232 @@
...
+                        on:click={() => {setUnsavedChanges(); deleteEntry(index)}}>
+                    Löschen
+                </Button>
+            {/each}
+        </div>
+        <div class="dashboard-admin-time-slots-button-wrapper">
+            <Button ariaLabel="Klicke, um einen neuen Time-Slot hinzuzufügen"
+                    on:click={() => {setUnsavedChanges(); addSlot(); }}>Hinzufügen
+            </Button>
+            <Button ariaLabel="Klicke, um änderungen zu speichern"
```
mgerhold, line 208:
```suggestion
            <Button ariaLabel="Klicke, um Änderungen zu speichern"
```

mgerhold (changes requested):

src/routes/dashboard/admin/time-slots/+page.svelte
PurpurTentakel97: This endpoint wasn't event there when I startet developing this.
But yes I can use it. :)

src/routes/dashboard/admin/time-slots/+page.svelte
PurpurTentakel97: Do you really think that this would help that much?
You would have to edit it ether way because event the calculated value will not be the right value. Last time we had a little break eg.

mgerhold: As I said, this is not a must. But in my browser, since it is set to English, all the datetime inputs use the 12-hour format and I have to set A.M. and P.M. for all times. But again, we should not waste time on a feature that we only use once every few months.

src/lookup/iconLookup.ts
```diff
@@ -22,6 +22,8 @@ const lookup: IconDict = {
     CircleCheck: 'fa-solid fa-circle-check',
     ArrowUp:     'fa-solid fa-arrow-up',
     ArrowDown:   'fa-solid fa-arrow-down',
+    Broadcast:   'fa-solid fa-tower-broadcast',
+    Video:         'fa-solid fa-video',
```
mgerhold, line 26: Formatting is a bit weird here.

src/routes/dashboard/admin/time-slots/+page.ts
```diff
@@ -1,23 +1,33 @@
...
     fetch: typeof globalThis.fetch
 }): Promise<LoadAdminTimeSlots> {
-    const allEventPromise = fetch(apiUrl('/api/dashboard/admin/all-events'));
-    const allEvents       = await checkAndParseInputDataAsync(
-        await allEventPromise,
-        dashboardAllEventsScheme,
-        `Serveranfrage für alle Events nicht erfolgreich. throw error(406)`,
-        `Unerwartete Daten für alle Events. throw error(406)`,
+    const allEventPromise     = fetch(apiUrl('/api/dashboard/admin/all-events'));
+    const talkDurationPromise = await fetch(apiUrl('/api/talk-duration-choices'));
```
mgerhold, line 12: Is the `await` here correct?

mgerhold (changes requested):
There's a little problem with the toggle icons. Look at this, for example:
image
The mouse cursor is a pointer, but when I click, nothing happens. I have to move the cursor closer to the icon to make it work:
image

The distance itself is not the problem, but the look of the cursor is not consistent with the behavior.

src/routes/dashboard/admin/time-slots/+page.svelte
PurpurTentakel97: okay

src/routes/dashboard/admin/time-slots/+page.ts
PurpurTentakel97: good call

PurpurTentakel97:
I removed the cursor altogether on the popup.
Is that better?

mgerhold:
> I removed the cursor altogether on the popup. Is that better?

Yes, I think it's better now. I'm just wondering why the area for the tooltip is different than the area for the click. But I think the current version is okay.

mgerhold (approved):

PR #137, opened 2025-01-18
User dashboard

PurpurTentakel97:
I just turn this into a PR so we can write here.

I think I need to rebase this after all.
But this will work.

### DIFF
--- a/src/routes/dashboard/user/profile/+page.svelte
+++ b/src/routes/dashboard/user/profile/+page.svelte
@@ -9,6 +9,11 @@
     import HeadlineH2 from 'elements/text/headlineH2.svelte';
     import Input from 'elements/input/input.svelte';
     import Button from 'elements/input/button.svelte';
+    import SaveMessage from 'elements/text/saveMessage.svelte';
+    import MessageWrapper from 'elements/text/messageWrapper.svelte';
+
+    let saveMessage: SaveMessage;
+    let errorMessages: string[] = [];
 
     let name: string         = '';
     let mail: string         = '';
@@ -36,6 +41,8 @@
 />
 
 <SectionDashboard classes="standard-dashboard-section dashboard-user-profile-section-wrapper">
+    <SaveMessage bind:this={saveMessage} />
+    <MessageWrapper messages={errorMessages} />
     <form class="dashboard-user-profile-section-entry-wrapper"
           on:submit|preventDefault={saveName}>
         <HeadlineH2>Name</HeadlineH2>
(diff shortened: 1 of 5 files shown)


src/elements/input/durationArray.svelte
mgerhold, line 1: Is it correct that this file is part of this PR? Or did you accidentially rebase onto the `speaker-talks` branch? Maybe rebase onto `main` first? This also applies to other files in this PR.

src/routes/dashboard/user/profile/+page.svelte
```diff
@@ -2,22 +2,266 @@
...
+                       type="password"
+                       bind:value={newPassword2}
+                       on:input={setUnsavedChanges} />
+                <Input id="dashboard-user-profile-input-password-old"
+                       labelText="Altes Passwort:"
+                       placeholderText="Altes Passwort"
+                       ariaLabel="Trage hier dein altes Passwort ein"
+                       type="password"
+                       bind:value={oldPassword}
+                       on:input={setUnsavedChanges} />
```
mgerhold, line 223: Can you put the old password on top, please? :)

mgerhold (changes requested):
As already commented on the one file, I think something got messed up here on this branch. The PR seems to also contain the other PR (or at least parts of it). Maybe you should try to clean that up first. My initial thoughts on how that would work:
- Do an interactive rebase on this branch, starting from the latest commit that also is on `main`. Drop all the commits that don't belong to this PR in the process.
- Rebase onto `main` (maybe doesn't really do anything, but that's okay).
- force-with-lease-Push again

Personally, I'd recommend doing the interactive rebase in the JetBrains IDE, because I think it's simpler to do than in the terminal (but maybe that's just me).

src/elements/input/durationArray.svelte
PurpurTentakel97: I tries to rebase onto main.
No changes.
Just ignore this PR until the other is merged and ich rebase again.

mgerhold: Okay 👍

src/routes/dashboard/user/profile/+page.svelte
PurpurTentakel97: Why would you do that?
The password is at all 3 cases the last field.
Why should that be different here.

Or should I move the password in all cases?

mgerhold: Because me and my chat immediately thought that this is the wrong order 😅 All websites I know put the old password to the top.
For the other two cases, it would not be that important to me. But maybe consistency would be good, so maybe put it to the top in all cases, if that's okay for you.

PurpurTentakel97: okay.
Than I change that now.

PurpurTentakel97:
Should I display the current data at the dashboard or should I remove that for now?

mgerhold (approved):

PR #138, opened 2025-01-24
Save

### LINKED ISSUE #135: wrong error message lookup when logging in with wrong password
The backend had a change a while ago, and now the frontend doesn't show the correct error message when trying to log in using a invalid password:

Image

### LINKED ISSUE #136: invalid social media URL during speaker application doesn't scroll to error message
When applying as a speaker and trying to use an invalid social media URL, the view doesn't automatically scroll up to the error message (and thus, the error message is not visible).

PurpurTentakel97:
This PR introduces:
- a single response lookup
- a new save function that parses errors when not successful
- a new wrapper save function that resets unsaved changes (I mostly use that one or I comment why I use the other one)
- a Message wrapper for multiple error messages.

Sorry for this rather big PR but I think its war worth it to have one save function and error parse function. 
Is it possible for you to check the hole page? I tried to cover it and I think its working. But it would be nice if you could make sure.
thx.

Is there a way to get all possible responses?
I tried to get all responses from previous PRs.
And I also add a response if a "new" one "appears".
But I feel like I'm missing some out.
They will not be displayed. There is only a error print and a placeholder error message.
But it would be nice to catch all responses somehow.  

close #136   
close #135


mgerhold (commented):
* I don't know whether it was caused by this PR, but in Chrome I now get a horizontal scrollbar on the mainpage (in Firefox it's okay, though).
image
* On the speaker application page, the view does not scroll up to the error messages (I don't know if you already told me that you're going to fix this or not 😅).

PurpurTentakel97:
Should be fixed by now.

src/routes/dashboard/user/application/+page.svelte
```diff
@@ -62,10 +63,11 @@
         }
         data.data.socials.socials = socialResult.data;
 
-        return errorQueue.length == 0;
+        return errorQueue.length === 0;
```
mgerhold, line 66:
```suggestion
        return errorQueue.length ==== 0;
```
(Kappa, obviously)

src/routes/dashboard/user/profile/+page.svelte
```diff
@@ -84,7 +88,12 @@
         const response = await save(toSave, '/api/account/change-email');
 
         if (response.success) {
-            mail = '';
+            mail       = '';
+            errorQueue = [ // assignment because of reactivity
+                ...errorQueue,
+                'Du musst deine Mail erst bestätigen damit sie hier angezeigt wird.',
```
mgerhold, line 94:
```suggestion
                'Du musst deine E-Mail-Adresse erst bestätigen, damit sie hier angezeigt wird.',
```

peddermaster2 (commented):
You are using `errorQueue` everywhere, but it it more like an `errorList` or `errorArray`.

PR #141, opened 2025-01-29
Apply talk

PurpurTentakel97:
this PR introduces a page where you can submit a talk.


src/menu/menuItems.ts
```diff
@@ -72,6 +72,11 @@ export const speakerSocialMedia: MenuItem = {
     url:         '/dashboard/speaker/socials',
     description: 'Klicke hier, um zu den Social-Media-Einstellungen zu gelangen',
 };
+export const speakerApplication: MenuItem = {
+    name:        'Talk einreichen',
```
mgerhold, line 76: I think this is good. But to make it easier to differentiate between the different tabs, maybe we should rename the "Talk" tab to "Meine Talks" or "Talks verwalten" or something like that?

src/pages/speakerTalkForm.svelte
```diff
@@ -0,0 +1,120 @@
...
+    import SaveMessage from 'elements/text/saveMessage.svelte';
+
+    export let classes: string = '';
+    export let data: DashboardPendingTalk;
+    export let tags: AllTalkTag;
+    export let talkDurations: DashboardTalkDurationChoices;
+
+    let errorList: string[] = [];
+    let saveMessage: SaveMessage;
+    let dispatcher          = createEventDispatcher();
```
mgerhold, line 29: Formatting is a bit weird here.

src/routes/dashboard/speaker/application/+page.svelte
```diff
@@ -0,0 +1,71 @@
...
+        id:                 0,
+        event_id:           data.event ? data.event.id : 0,
+        title:              '',
+        description:        '',
+        tags:               [],
+        possible_durations: [],
+        notes:              '',
+        requested_changes:  '',
+
+    };
```
mgerhold, line 28:
```suggestion
        requested_changes:  '',
    };
```

src/routes/dashboard/speaker/application/+page.svelte
```diff
@@ -0,0 +1,71 @@
...
+        <div class="dashboard-speaker-application-success-wrapper">
+            <TextLine>Dein Talk wurde erfolgreich eingereicht.</TextLine>
+            <TextLine>Du bekommst jetzt eine E-Mail.</TextLine>
+            <TextLine>Sobald wir deinen Talk bearbeitet haben, werden wir dich benachrichtigen.</TextLine>
+            <TextLine>Du möchtest deinen Talk noch bearbeiten? Kein Problem. Er wird bereits im Talk-Tab angezeigt.
+            </TextLine>
+            <StyledLink text={MenuItem.speakerTalk.name}
+                        href={MenuItem.speakerTalk.url}
+                        title={MenuItem.speakerTalk.description}
+                        newTab={false} />
```
mgerhold, line 46: In general, I think it's a bit redundant to have such a link here. The "real" link is directly above the text. Having two ways of navigating there may be confusing and the speaker may not understand the "regular" way of going there. Also, it looks a bit weird in that place IMHO. This is the same like on the page where one can apply as a speaker: If the application was successful, buttons appear that serve as a secondary (i.e. redundant) navigation. I'm not a fan of that, since I think there should be one unique way of navigating. But what do you think?

src/pages/speakerTalkForm.svelte
PurpurTentakel97: This is what the formatter is doing. 😅

mgerhold: Then please tell the formatter to go home. It is drunk 😂
But no, then ignore it, it's okay :)

src/routes/dashboard/speaker/application/+page.svelte
PurpurTentakel97: I can remove all that.
But I think it not that much of an issue.
This will only be displayed when you submitted a talk.
How would you use that as a navigation?
Always submit a new talk to get to the other tab? That sems a bit off.

But I can remove the buttons at the speaker application because  those are visible all the time because once you're a speaker you can not apply again. And than just add a text that says basically "Just navigate yourself".

mgerhold: Yes, exactly, that's what I mean: Let's just remove the redundancy here, so that there's only one way of doing things :)

mgerhold: I would remove them all. As I said (and as you also said), the button is only visible after the talk has been submitted. The user may see the button and click it, but doesn't understand that there's another way of navigating there (directly above it).
I mean, if you don't want to remove it, we should not discuss too long about it. This is just my opinion, but I can also live with it, if you don't remove it. Your choice :)

PurpurTentakel97: No problem.

I will remove all buttons whatsoever.

mgerhold: Okay, thanks. I hope you don't feel forced to do that 🙈

mgerhold (approved):

PR #142, opened 2025-01-30
Event

PurpurTentakel97:
I had to split the admin event page because this combined version courses some bugs I can't fix. 

So there is a new Tab at the admin dashboard right now where a new event can be submitted.
Functionality wise it offers exactly the same features a before just in two separate pages. 

Hope this is not to bad. 😅


mgerhold (approved):
Seems to work well q(≧▽≦q)

PR #145, opened 2025-02-01
Reject

### LINKED ISSUE #144: some images are not shown
As you already recognized in your stream, since the latest backend changes some of the images are not showing anymore. However, I think this is a frontend but. For example here:

https://github.com/TechStreamConference/website-frontend/blob/main/src/routes/dashboard/admin/approval/%2Bpage.svelte#L181

You're using `apiUrl()` to fetch the images' URL, but I think this should be `imageUrl()` instead. Because of this, the `/images` part of the URL is missing and thus the images cannot be found.

PurpurTentakel97:
this PR introduces a new approval page.
This page is sorted by `user_id`.

It uses borders to make clear with data belong to witch entry.
The new page seams to be al lot cleaner than the old page.
im not quite sure if I like all the borders.
I would keep the border around a user but one could remove the border around each entry. 
just try to remove `speaker-team-member-wrapper` from a speaker / team member and the  'social-media-approval-wrapper' from a social media link to try out how that would look without that extra border. 

Also:
I fixed the big margin at the speaker approval page
close #144


src/elements/text/subHeadline.svelte
```diff
@@ -1,7 +1,7 @@
 <script lang="ts">
-    export let classes: string = '';
-    export let size: string    = 'sub-headline-big';
-    export let id: string      = '';
+    export let classes: string                                 = '';
+    export let size: 'sub-headline-small' | 'sub-headline-big' = 'sub-headline-big';
```
mgerhold, line 3: Is this Typescripts way of specifying an enum? monkaS

src/routes/dashboard/admin/approval-speaker-team-member/+page.svelte
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    function cleanUpSocialMediaLinks(user_id: number): void {
+        setTimeout(() => {
+            data.socialMedia = data.socialMedia.filter((socialMedia) => {
+                return socialMedia.user_id !== user_id;
+            });
+
+            data.userIDArray = getUserIds(data.speaker, data.teamMember, data.socialMedia);
+        }, 2900); // move this timeout a litte bit forward to make sure the social media links gets cleanup bevor delete of speaker / team member
```
mgerhold, line 87:
```suggestion
        }, 2900); // move this timeout a litte bit forward to make sure the social media links gets cleaned up before deletion of speaker / team member
```

mgerhold (approved):

PR #150, opened 2025-02-06
Talk approval

PurpurTentakel97:
This PR introduces that same approval functionality as the speaker / team member / social media link approval page just for talks.

Note: I use the `user_id` here as a heading because I don't have the username of the speaker.
Do you have an Idea how to better do that?


mgerhold:
> Note: I use the `user_id` here as a heading because I don't have the username of the speaker.
> Do you have an Idea how to better do that?

You should just be able to use the speaker name for the heading instead, since the speaker is uniquely identifier by the combination of user id and event id. However, it may be a problem that you don't consider the event id for this page. But it would only be a concern if we would have multiple events opened at the same time (which, I think, is rather unrealistic).
tl;dr: Just use the speaker name for the heading and everything should be fine.

src/routes/dashboard/admin/approval-talk/approvalHelper.ts
```diff
@@ -0,0 +1,65 @@
...
+    const ids = new Set<number>();
+
+    pendingTalks.forEach(talk => ids.add(talk.speaker.user_id));
+    tentativeTalk.forEach(talk => ids.add(talk.speaker.user_id));
+
+    return [...ids].sort((a, b) => a - b);
+}
+
+export async function fetchTentativeTalks(fetch: typeof globalThis.fetch): Promise<DashboardAllTentativeOrAcceptedTalks> {
+    // TODO replace tentative talks fetch
```
mgerhold, line 28: Why this todo? Isn't this the new fetch already? I think I'm missing something here.

src/routes/dashboard/admin/approval-talk/+page.ts
```diff
@@ -0,0 +1,89 @@
...
+    return {
+        pendingTalks:   await pendingTalksParsePromise,
+        tentativeTalks: await tentativeTalksPromise,
+        userIDArray:    getUserIds(await pendingTalksParsePromise, await tentativeTalksPromise),
+        slots:          await fetchTentativeSlots(fetch, await tentativeTalksPromise),
+    };
+}
+
+
+// request feature: route for only admin all tentative talks
```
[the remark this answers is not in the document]
PurpurTentakel97: I just not remove this comment. 🙈

src/routes/dashboard/admin/approval-talk/approvalHelper.ts
PurpurTentakel97: I just not remove this comment. 🙈

PurpurTentakel97:
> I think that talks don't become visible after rejecting them.
What do you mean by this.
How should a talk become visible after rejecting. 😅

mgerhold:
> > I think that talks don't become visible after rejecting them.
> 
> What do you mean by this. How should a talk become visible after rejecting. 😅

Sorry, I meant "invisible" 🙈

PurpurTentakel97:
>>> I think that talks don't become visible after rejecting them.
>>
>>What do you mean by this. How should a talk become visible after rejecting. 😅
>
>Sorry, I meant "invisible" 🙈

But this is already happening. 
What did you do that this is not working on your setup?

PurpurTentakel97:
> Also, after approving a talk, the page should refresh, so that the talk reappears as a tentative talk.

This was already implementing. But there was a bug because the slots were not fetched when there was no tentative talks.
I not fetch the slots based on tentative and pending talks.
This should fix it.

mgerhold:
> But this is already happening. What did you do that this is not working on your setup?

I tested it again at it worked. Maybe I just overlooked that before. So I guess it is okay then 🤔

PurpurTentakel97:
>You should just be able to use the speaker name for the heading instead, since the speaker is uniquely identified by the combination of user id and event id. However, it may be a problem that you don't consider the event id for this page. But it would only be a concern if we would have multiple events opened at the same time (which, I think, is rather unrealistic).
tl;dr: Just use the speaker name for the heading and everything should be fine.

Yes. I think I dont have the event names now.
I would have to fetch the event Data then to get all event names.
This is not verry complex but I also think it is not necessary.

PurpurTentakel97:
> When assigning a time slot, the dropdown contains all possible time slots, not only those that are still free. I have created https://github.com/TechStreamConference/website-backend/issues/160 in the backend to extend the time slots endpoint so that it contains this information. Then, you will be able to filter only valid time slots (or add an information to each dropdown entry).

I added a issue as well so that we don't have to block the PR with that.
Filtering should not be the Problem right here. (#151)

mgerhold (approved):

PR #156, opened 2025-02-16
Panel discussion

### LINKED ISSUE #148: Implement panel discussions
implement panel discussions at frontpage and provide a create panel discussion admin page

PurpurTentakel97:
this PR introduces 2 parts.
first: displaying guests at the frontpage
second: provide guests for talks.

I noticed, that one can not remove all guests from a talk because the backend does not support this. It returns
```json
{
    "guest_ids.*": "The guest_ids.* field is required."
}
```
when the payload looks like this: `{"guest_ids":[]}`

Is that a problem?

close #148 
new #155 (witch means the page were one can provide a new panel.

### DIFF
--- a/src/elements/input/personArray.svelte
+++ b/src/elements/input/personArray.svelte
@@ -1,12 +1,14 @@
 <script lang="ts">
-    import Label from 'elements/text/label.svelte';
     import type { DashboardAllPersons, DashboardPerson } from 'types/dashboardProvideTypes';
+
+    import { createEventDispatcher } from 'svelte';
+    import { imageUrl } from 'helper/links';
+
+    import Label from 'elements/text/label.svelte';
     import Tag from 'elements/text/tag.svelte';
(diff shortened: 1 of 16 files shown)


mgerhold:
> Is that a problem?

I could provide an additional endpoint to remove all guests from a talk. If I would allow an empty array in the existing endpoint, I would have to do the validation myself and could not use CodeIgniters validators. Do you want me to add such an endpoint?

PurpurTentakel97:
> I could provide an additional endpoint to remove all guests from a talk. If I would allow an empty array in the existing endpoint, I would have to do the validation myself and could not use CodeIgniters validators. Do you want me to add such an endpoint?

Ah I see.
The question is: Do we need this feature.
I just did mention that this is not possible.
But if we think that not a problem.
Im fine with that.

mgerhold (approved):

PR #157, opened 2025-02-17
New panel discussion

### LINKED ISSUE #155: implement new panel discussion

PurpurTentakel97:
Now you can apply a new panel discussion at the admin panel.
note: Im using the talk template for a Speaker Application here.
This results in a unused field 'notes'.
I think this is ok because it does not affect the workflow at it is only appearing at the admin panel.
So I don't see writing a new form here.

close #155


mgerhold (approved):
Looks good, works fine! 👍

PR #158, opened 2025-02-19
Reload by same link

### LINKED ISSUE #133: site should refresh when navigating to the current route
When you're on some page of the website and click a link that leads you to the same page again, the page contents do not reaload, but they should. For example, when being on the "Talk" tab in the speaker dashboard, making changes, and then clicking the "Talk" tab, you're being asked whether you really want to leave the current page, but after confirming, the contents of the site stay the same (instead they should be reloaded).

PurpurTentakel97:
I think I managed to reload the page when the URL is the same.
Everything seems to work just fine after all.
But maybe you can check that. 

close #133

### DIFF
--- a/src/routes/dashboard/admin/new-event/+page.svelte
+++ b/src/routes/dashboard/admin/new-event/+page.svelte
@@ -42,7 +42,7 @@
       entryName={MenuItem.adminNewEvent.name}
       classes="navigation-tabs-dashboard-subpage" />
 
-<SectionDashboard classes="dashboard-admin-event-section">
+<SectionDashboard classes="standard-dashboard-section">
     {#if currentState === State.Saved}
         <div class="dashboard-admin-new-event-message-wrapper">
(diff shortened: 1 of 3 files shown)


src/routes/dashboard/admin/new-event/+page.svelte
```diff
@@ -42,7 +42,7 @@
       entryName={MenuItem.adminNewEvent.name}
       classes="navigation-tabs-dashboard-subpage" />
 
-<SectionDashboard classes="dashboard-admin-event-section">
```
PurpurTentakel97, line 45: Actually.
I have no Idea how that worked before.
This class is not even existing.
But for some reason svelte did apply the right class most of the time.
I like web......

mgerhold: Web stuff...

PurpurTentakel97:
Close this because of web is doing web things

PR #160, opened 2025-02-22
Contact

PurpurTentakel97:
this introduces a contact page

### DIFF
--- a/src/elements/input/input.svelte
+++ b/src/elements/input/input.svelte
@@ -1,7 +1,12 @@
 <script lang="ts">
+    import type { Color } from 'helper/basic';
+
     import { typeWorkaround } from 'types/workaround';
+    import { lerpColor, orange } from 'helper/basic';
+    import { onMount } from 'svelte';
 
     import Label from 'elements/text/label.svelte';
+    import TextLine from 'elements/text/textLine.svelte';
 
     export let ariaLabel: string;
 
@@ -11,58 +16,106 @@
     export let labelText: string       = '';
     export let placeholderText: string = '';
     export let fileAccept: string      = '';
+    export let limit: number           = -1;
 
     export let value: string | number | File = '';
+    let colorString: string                  = '';
 
     let input: HTMLInputElement;
+
+    onMount(() => {
+        calcColor();
+    });
 
     export function clear(): void {
         input.value = '';
     }
+
+    function calcColor(): void {
+        if (limit === -1) {
+            return;
+        }
+        if (typeof value !== 'string' && typeof value !== 'number') {
+            return;
+        }
+
+        const length = typeof value === 'string' ? value.length : value.toString().length;
+        let factor   = 1 - (length / limit);
+        if (factor > 0.5) {
+            factor = 1;
+        } else {
+            factor *= 2;
+        }
+
+        const color: Color = lerpColor(orange, {
+            red:   255,
+            green: 255,
+            blue:  255,
+        }, factor);
+        colorString        = `rgb(${color.red},${color.green}, ${color.blue})`;
+    }
 </script>
 
-<div class={classes}>
+<div class="{classes} input-line-wrapper">
     <Label for_={id}>{labelText}</Label>
-    {#if type === 'file'}
-        <input
-              bind:this={input}
-              class="normal-font"
-              {id}
-              type="file"
-              accept={fileAccept}
-              name={id}
-              placeholder={placeholderText}
-              bind:value
-              on:input
-              on:change
-              aria-label={ariaLabel}
-              step="1"
-        />
-    {:else}
-        <input
-              bind:this={input}
-              class="normal-font"
-              {id}
-              use:typeWorkaround={type}
-              name={id}
-              placeholder={placeholderText}
-              bind:value
-              on:input
-              on:change
-              aria-label={ariaLabel}
-              step="1"
-        />
-    {/if}
+    <div class="input-line-entry-wrapper">
+        {#if type === 'file'}
+            <input
+                  bind:this={input}
+                  class="normal-font input-line"
+                  {id}
+                  type="file"
+                  accept={fileAccept}
+                  name={id}
+                  placeholder={placeholderText}
+                  bind:value
+                  on:input
+                  on:change
+                  aria-label={ariaLabel}
+                  step="1"
+            />
+        {:else}
+            <input
+                  bind:this={input}
+                  class="normal-font input-line"
+                  {id}
+                  use:typeWorkaround={type}
+                  maxlength={limit}
+                  name={id}
+                  placeholder={placeholderText}
+                  bind:value
+                  on:input={calcColor}
+                  on:change
+                  aria-label={ariaLabel}
+                  step="1"
+            />
+        {/if}
+        {#if limit !== -1}
+            {#if typeof value === 'string'}
+                <TextLine classes="input-limit-text text-color-custom"
+                          --color={colorString}>{value.length} / {limit}</TextLine>
+            {:else if typeof value === 'number'}
+                <TextLine classes="input-limit-text text-color-custom"
+                          --color={colorString}>{value.toString().length} / {limit}</TextLine>
+            {/if}
+        {/if}
+    </div>
 </div>
 
 <style>
-    div {
+    .input-line-wrapper {
         display:        flex;
         flex-direction: column;
         width:          100%;
     }
 
-    input {
+    .input-line-entry-wrapper {
+        display:        flex;
+        flex-direction: row;
+        position:       relative;
+    }
+
+    .input-line {
         color:            var(--white-color);
         background-color: var(--primary-color-light);
         border-radius:    var(--border-radius);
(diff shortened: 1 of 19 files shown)


src/elements/input/input.svelte
```diff
@@ -11,22 +17,64 @@
     export let labelText: string       = '';
     export let placeholderText: string = '';
     export let fileAccept: string      = '';
+    export let limit: number           = -1;
```
mgerhold, line 20: Using a "magical sentinel value" (like -1) is usually considered bad practice. But I don't know about web stuff 😅
Couldn't you just use a nullable type instead?

src/elements/input/textArea.svelte
```diff
@@ -9,27 +14,60 @@
     export let classes: string         = '';
     export let placeholderText: string = '';
     export let rows: number            = 10;
+    export let limit: number           = -1;
```
mgerhold, line 17: Same question here

src/elements/input/textArea.svelte
```diff
@@ -9,27 +14,60 @@
...
-    export let value: string | number = '';
+    export let value: string = '';
+    let colorString: string  = '';
 
     const dispatch = createEventDispatcher();
 
+    onMount(() => {
+        if (limit === -1) {
+            return;
+        }
```
mgerhold, line 27: In the input component, you do this check inside the `calcColor()` function, but here you do it outside. Maybe harmonize that?

src/elements/input/textArea.svelte
```diff
@@ -9,27 +14,60 @@
...
+            factor *= 2;
+        }
+
+        const color: Color = lerpColor(orange, {
+            red:   255,
+            green: 255,
+            blue:  255,
+        }, factor);
+        colorString        = `rgb(${color.red},${color.green}, ${color.blue})`;
+    }
```
mgerhold, line 53: This code is basically duplicated. Can you extract it into the helpers? I mean, I know that it modifies a global variable, but maybe you can refactor the function so that it only `return`s the color string. Then, you could wrap a call to this function inside a lambda. This would remove the duplication and keep the code both smaller and simpler.

src/routes/contact/+page.svelte
```diff
@@ -0,0 +1,182 @@
...
+    import Button from 'elements/input/button.svelte';
+    import SaveMessage from 'elements/text/saveMessage.svelte';
+    import MessageWrapper from 'elements/text/messageWrapper.svelte';
+    import UnsavedChangesPopup from 'elements/popups/unsavedChangesPopup.svelte';
+    import UnsavedChangesCallbackWrapper from 'elements/navigation/unsavedChangesCallbackWrapper.svelte';
+    import { onMount } from 'svelte';
+
+    enum State {
+        Default,
+        EmailSend,
```
mgerhold, line 27: I think this should be in the past form.
```suggestion
        EmailSent,
```

src/routes/contact/+page.svelte
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        if (message.trim().length === 0) {
+            errorList.push('Bitte gibt eine Nachricht ein.');
+        }
+
+        return errorList.length === 0;
+    }
+
+    async function save(): Promise<boolean> {
```
mgerhold, line 83: Aren't you using the `Async`-suffix for `async` functions?

src/routes/contact/+page.svelte
```diff
@@ -0,0 +1,182 @@
...
+                <div class="contact-subject-wrapper">
+                    <DropDown id="contact-subject-dropdown"
+                              labelText="Betreff:"
+                              data={subjectData}
+                              bind:selected={subject} />
+                    {#if subject === subjectOther}
+                        <Input id="contact-subject-concrete-input"
+                               labelText="Betreff:"
+                               placeholderText="Betreff"
+                               ariaLabel="Spezifiziere hier dein Betreff"
```
mgerhold, line 141:
```suggestion
                               ariaLabel="Spezifiziere hier den Betreff"
```
(or `deinen`)

mgerhold (changes requested):
* Would it be possible to have the "Name" field being focussed when the page opens? I think at the moment "Betreff" has the focus.

* I would suggest making the text area way smaller. At the moment, the complete contact form (at default size) does not fit in my browser window (at 100 % zoom). To be honest, I'm not even sure if we should make it resizable at all. It looks a bit weird to move the footer around 😅

* When being on the Contact page, I think that "Kontakt" should still be visible in the footer navigation. In general, I think that the navigation items there should not change, regardless of the page you are on.

* There's no auto-scroll to the top when error messages appear.

* When I enter the following (incomplete) data, an unknown error apppears, see screenshot:
  
image

* Can we rename the first "Betreff" to "Kategorie" or "Worum geht's?"? When chosing "Sonstiges", the heading "Betreff" is duplicated which looks weird.

* After successfully sending a message, when I then press F5, the old data is visible for a fraction of a second. If this is a Svelte problem, maybe just clear all data after successful sending?

* When I enter valid data and then hit the send button, there's an error message, but it's only visible for a very short time. After that time, the success message becomes visible.

PurpurTentakel97:
>Would it be possible to have the "Name" field being focussed when the page opens? I think at the moment "Betreff" has the focus.

I added this issue: #161

PurpurTentakel97:
>I would suggest making the text area way smaller. At the moment, the complete contact form (at default size) does not fit in my browser window (at 100 % zoom). To be honest, I'm not even sure if we should make it resizable at all. It looks a bit weird to move the footer around 😅

I can make it smaller.
We should definitely make it resizable.
The one thing I hate the most at contact forms is that I can not resize this stuff.
To have a  confutable area to write in.

PurpurTentakel97:
>When being on the Contact page, I think that "Kontakt" should still be visible in the footer navigation. In general, I think that the navigation items there should not change, regardless of the page you are on.

Its a litte late for that comment.
Now all footer are separated.
I mean I can refactor that and make it a single file.
But do we need this.

And why so you don't want 'Kontakt' at the footer on the contact page.
When I add it here I will change all footer so stay consistent.

PurpurTentakel97:
>When I enter the following (incomplete) data, an unknown error apppears, see screenshot:1

What is the key, that is missing?

PurpurTentakel97:
>When I enter valid data and then hit the send button, there's an error message, but it's only visible for a very short time. After that time, the success message becomes visible.

I am not able to reproduce that.
Is that every time?

src/elements/input/input.svelte
PurpurTentakel97: Would it be okay to define a variable here?
Something like `const defaultLimit = -1` and then use this?
Because this one guy in my chat today said that -1 would be a common value do a non existing limit.

mgerhold:
> The question is: How many lines fit on your screen?

On my screen, it looks like this:
image

So there are already plenty of lines. But I think a default of 6 lines or so should be sufficient.

> Its a litte late for that comment.
Now all footer are separated.
I mean I can refactor that and make it a single file.
But do we need this.

If its too much work, then we should not do it, obviously. But if it reduces the amount of code, maybe it would be a good thing.

> And why so you don't want 'Kontakt' at the footer on the contact page.
When I add it here I will change all footer so stay consistent.

In my case, I just wanted to write a second message after the first one, but the navigation link was gone. Sure, I can hit F5, but that's not obvious, since it's not immediately clear whether the success page is the same page as the contact page, or a different one. And yes, in practice, noone will ever write two messages in a row. But isn't this just a common pattern on most websites that the footer navigation always stays the same? I think so, at least.

> What is the key, that is missing?

The response of the backend is this:
```json
{
    "email": "The email field must contain a valid email address.",
    "subject": "The subject field is required.",
    "message": "The message field is required."
}
```
So I guess the problem is the subject. But it's weird that you send the request at all when validation fails. But I think you already identified that problem.

> I am not able to reproduce that.
Is that every time?

I think it only happens the first time (or after using `Ctrl` + `F5`).

src/elements/input/input.svelte
mgerhold: Sure, -1 is a common value in many programming languages. But in my eyes, this style of doing it is outdated. It's not that big of a problem, however, so just decide for yourself :)

PurpurTentakel97:
I added this issue: #164

PurpurTentakel97:
I hope I did not miss anything.
this conversation was a little confusing to me. 😅

mgerhold:
I doesn't seem to work anymore, because the validation of the email address always fails:
image

src/elements/input/input.svelte
```diff
@@ -17,7 +15,8 @@
     export let labelText: string       = '';
     export let placeholderText: string = '';
     export let fileAccept: string      = '';
-    export let limit: number           = -1;
+    export const NO_LIMIT: number      = -1;
+    export let limit: number           = NO_LIMIT;
```
mgerhold, line 19: theres-no-limit-2unlimited

src/routes/contact/+page.svelte
```diff
@@ -131,14 +141,14 @@
                        on:input={setUnsavedChanges} />
                 <div class="contact-subject-wrapper">
                     <DropDown id="contact-subject-dropdown"
-                              labelText="Betreff:"
+                              labelText="Was gibts?"
```
mgerhold, line 144:
```suggestion
                              labelText="Was gibt's?"
```

mgerhold (changes requested):
Please see my other comment regarding the failing email validation.

mgerhold (approved):

PR #168, opened 2025-03-03
Focus

### LINKED ISSUE #161: initial focus on all form pages

PurpurTentakel97:
close #161 

I can focus DOM elements.
I talked with Arti about initial focus.
He said that I should be verry careful with that because for screen readers it seams so be normal to step threw the menu each time. That way they can unterstand the page. 
So I added the initial focus only on pages where you navigate only to enter something. eg. login page.

We can discuss if it is missing at some points.

What do you think?

### DIFF
--- a/src/helper/basic.ts
+++ b/src/helper/basic.ts
@@ -62,3 +62,14 @@ export function trimOrNull(entry: string): string | null {
     }
     return trimmed;
 }
+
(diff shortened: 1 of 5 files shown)


mgerhold (approved):
Looks good to me and seems to work well 👍

PR #169, opened 2025-03-12
feat: flesh out explanation texts

mgerhold:
Please note that this pull request does not merge into `main`, but rather into `106-explanation-texts`.

My formatter seems to work differently than yours (even though it's also WebStorm 🤷). Maybe you'll have to reformat the file on your end to end up with your prefered style.

### DIFF
--- a/src/forms/speakerTalkForm.svelte
+++ b/src/forms/speakerTalkForm.svelte
@@ -1,13 +1,13 @@
 <script lang="ts">
-    import type { DashboardPendingTalk, DashboardTalkDurationChoices } from 'types/dashboardProvideTypes';
-    import type { SetTalk } from 'types/dashboardSetTypes';
-    import type { AllTalkTag } from 'types/provideTypes';
+    import type {DashboardPendingTalk, DashboardTalkDurationChoices} from 'types/dashboardProvideTypes';
+    import type {SetTalk} from 'types/dashboardSetTypes';
+    import type {AllTalkTag} from 'types/provideTypes';
 
-    import { setUnsavedChanges } from 'stores/saved';
-    import { scrollToAnchor } from 'helper/scroll';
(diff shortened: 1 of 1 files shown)


mgerhold:
Oh, and feel free to merge this PR yourself if you're satisfied with it :)

src/forms/speakerTalkForm.svelte
```diff
@@ -65,54 +66,73 @@
...
                   ariaLabel="Gib hier die Beschreibung des Talks ein"
                   bind:value={data.description}
                   on:input={setUnsavedChanges}
-                  on:submit={save} />
-        <Paragraph classes="paragraph-gray">Beschreibung</Paragraph>
-        <div class="grid-line" />
+                  on:submit={save}/>
+        <Paragraph classes="paragraph-gray">Beschreibe deinen Vortrag. Hier darfst du gerne etwas mehr ins Detail gehen.
+            Worum geht es in deinem Vortrag? Was wird man dabei lernen? Was ist das Besondere an deinem Vortrag? Eine
+            Länge von ca. 30 bis 50 Wörtern ist ideal.
```
PurpurTentakel97, line 97: Should I limit this than?
maybe not at 50 words but maybe at 100 words or something?
(I would need to calculate a letter count out of the words here)

PurpurTentakel97 (approved):

PurpurTentakel97:
In general one can discuss if we should limit all / some of the fields.

src/forms/speakerTalkForm.svelte
mgerhold: I was also thinking about that. Feel free to add it. Shouldn't be that hard (I think something like `message.remove(".,!:;()").split(" ").count()` or so should be sufficient). But I don't think it's a must-have.

PR #175, opened 2025-03-18
Vdo

PurpurTentakel97:
this introduces links at the admin and speaker dashboard

### DIFF
--- a/src/routes/dashboard/speaker/events/+page.svelte
+++ b/src/routes/dashboard/speaker/events/+page.svelte
@@ -5,10 +5,55 @@
     import type { LoadDashboard, LoadSpeakerTeamMemberEvent } from 'types/dashboardLoadTypes';
 
     import SpeakerTeamMemberEvent from 'pages/speakerTeamMemberEvent.svelte';
+    import SubHeadline from 'elements/text/subHeadline.svelte';
+    import VdoGrid from 'elements/vdo/vdo-grid.svelte';
+    import Button from 'elements/input/button.svelte';
+    import GeneralPopup from 'elements/popups/generalPopup.svelte';
+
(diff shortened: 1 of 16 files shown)


src/routes/dashboard/speaker/events/+page.svelte
```diff
@@ -3,16 +3,94 @@
...
+        <SubHeadline>Video Links des aktuellen Events:</SubHeadline>
+        {#if displayLinks}
+            <VdoGrid entries={vdoData} />
+        {:else}
+            <Button ariaLabel="Klicke hier, um deine aktuellen Video Links anzuzeigen"
+                    on:click={ () => {showLinkPopup.show(""); }}>Links anzeigen
+            </Button>
+        {/if}
+    </div>
+{/if}
```
mgerhold, line 75: I was expecting this to be shown on all the sub-pages under the "Speaker" tab, not only when "Events" is selected. That was the reason why I wanted this to be above the sub-navigation (where it already is). The positioning only makes sense if it is always visible.
So maybe you could just encapsulate this in another Svelte component and insert that component into all these pages ("Events", "Meine Talks", "Talk einreichen", and "Social Media"). Or maybe it would be better to put it into `src/routes/dashboard/speaker/+layout.svelte` (if that's possible).

mgerhold (changes requested):

src/routes/dashboard/speaker/events/+page.svelte
PurpurTentakel97: done

mgerhold (approved):
Looks good 👍

I think we'll have to add some explanatory text still. But we can do that separately later :)

PR #194, opened 2025-03-23
fix: rainbow tags

### LINKED ISSUE #180: [BUG] change colors in tag lookup
backend changes the color tags.
need to adapt that.
see: https://github.com/TechStreamConference/website-backend/pull/178

PurpurTentakel97:
close #180 

I added new colors for those tags.
Note that the colors are repeating to get rid of the rainbow effect.
I tried to separate all colors with a lot of tags.

obviously the formatter did things to.

### DIFF
--- a/src/lookup/tagColorLookup.ts
+++ b/src/lookup/tagColorLookup.ts
@@ -1,30 +1,39 @@
-
+// @formatter:off
 const colors: string[] = [
-    "#FF5733", "#33FF57", "#3357FF", "#FF33A6", "#33FFF6",
-    "#FFC300", "#DAF7A6", "#581845", "#C70039", "#900C3F",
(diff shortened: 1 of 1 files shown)


mgerhold (commented):
For my personal taste the colors are too "flashy"/"intense". I also did a poll on Twitch:
image

Especially in dark mode. Maybe it's possible to reduce the saturation component of all colors? You could even do this inside the code for testing. Or we have to ask a web designer to help us here 😅

PurpurTentakel97:
maybe I'll ask Propz.
I had less flashy colors before but I never get that worked with accessibility in terms of contrast. 😅

mgerhold:
Yes, let's ask someone who has more experience 😅

PR #216, opened 2025-04-26
Subgrid

### LINKED ISSUE #211: subgrid for person
check out subgrid for person rectangles to be able to reduce the space between subheadline and line.

PurpurTentakel97:
close #211 

This uses `subgrid` for the person rectangle.
This allows for a dynamic layout but also a layout where all grid elements of each row have the same layout.

### DIFF
--- a/src/elements/person/personGrid.svelte
+++ b/src/elements/person/personGrid.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-    import type { Person } from 'types/provideTypes';
+    import type {Person} from 'types/provideTypes';
 
     import PersonRectangle from './personRectangle.svelte';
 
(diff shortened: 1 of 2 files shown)


mgerhold:
The distances between the portraits and the name of the speakers seem to differ between different rows:
image

It seems like this only happens with Firefox, though 🙈

mgerhold (changes requested):
The distances between the portraits and the name of the speakers seem to differ between different rows:
image

It seems like this only happens with Firefox, though 🙈

PurpurTentakel97:
Firefox is doing a pretty good job in being MSVC of the web world. 🙈

mgerhold (approved):
As you told me, this bug will disappear as soon as we fix the photo dimensions. So we can merge it already.

PR #218, opened 2025-05-27
feat: add "vergeben" tag to time slots

### LINKED ISSUE #151: Filter Slots
Filter Time Slots for Talk approval when Backend implemented this.

PurpurTentakel97:
close #151

at the dropdown where you suggest the timeslot a "vergeben" is added to each occupied slot.
as you can see is not best practice that `is_occupied` can be undefined.
I set it myself to false now when it is undefined.

We can leave it like this. This will work.
It would be cleaner when you would set the value always and I just check it.
We can discuss that.

Do we need that `is_occupied` flag at an other page two?
Im not quite sure.

### DIFF
--- a/src/forms/adminTentativeTalkApprovalForm.svelte
+++ b/src/forms/adminTentativeTalkApprovalForm.svelte
@@ -87,6 +87,7 @@
         errorList = result.messages;
         if (result.success) {
             talk.suggested_time_slot = getElementByID(slots[talk.event_id], id);
+            talk.suggested_time_slot.is_occupied = true;
             dispatch('suggest');
         }
     }
--- a/src/forms/adminTentativeTalkApprovalForm.svelte
+++ b/src/forms/adminTentativeTalkApprovalForm.svelte
@@ -86,6 +86,7 @@
         message.setSaveMessage(result.success ? SaveMessageType.Save : SaveMessageType.Error);
         errorList = result.messages;
         if (result.success) {
+            getElementByID(slots[talk.event_id], talk.suggested_time_slot.id).is_occupied = false;
             talk.suggested_time_slot = getElementByID(slots[talk.event_id], id);
             talk.suggested_time_slot.is_occupied = true;
             dispatch('suggest');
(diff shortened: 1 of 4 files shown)


mgerhold:
When suggestion a time slot for a talk, the dropdown list does not update immediately. But after refreshing the site, it updates. This is not that big of a problem (because it's only in the admin dashboard), but if it's easy to fix, we should do it maybe.

PurpurTentakel97:
> I don't quite understand:
> 
> > as you can see is not best practice that `is_occupied` can be undefined.
> 
> According to the implementation in the backend, it should always be defined. Did you maybe switch back to another branch and then tested it? Of course, on the other branches it's missing (for now).

Ah I know why this is occurring. 
A talk is holding its time slot. And this time slot has the same zod scheme.
That slot is missing `is_occupied` then.
So the your new endpoint is just working find. I did get that wrong.

But when I add `is_occupied` to the slot scheme and a talk is provided, zod gets angry.
My solution now addresses that but it also will not trigger anymore if   `is_occupied`  would be undefined at the Drop Downs for some reason. Undefined will now always occur as false. If youre fine with that, I'm fine with that.

If not ether you have to provide this bool always for a time slot (across all endpoints) or I have to add a extra zod scheme just for the Drop Downs. 

What do you think?

PurpurTentakel97:
> > When suggestion a time slot for a talk, the dropdown list does not update immediately. But after refreshing the site, it updates. This is not that big of a problem (because it's only in the admin dashboard), but if it's easy to fix, we should do it maybe.
> 
> Its not that easy than I thought. I would add a callback oder a two way bind across three Components to update this for all Drop Downs. Should I do that or do we just leave it like this?

Tries id again and is was easier than I thought. LUL

mgerhold:
> Tried it again and is was easier than I thought. LUL

Now there's still another little bug 😅 When you suggest a time slot for a talk that already had another suggested time slot, then both time slots will appear as occupied until the page is refreshed 😅
I mean, this is not a big deal. Just tell me whether you want to fix that or if we merge it just like it is now.

mgerhold:
Oh...and the other thing (regarding the zod scheme): That's okay for me. No need to change that, I guess.

mgerhold (approved):

PR #220, opened 2025-06-16
Video

### LINKED ISSUE #210: self-hosted Trailer Video
@pr0pz pointed out that YouTube is the only source with DSGVO Cookie.
If we get rid of embedding YouTube we could get rid of a Cookie banner at all. 
So I suggest to self-host the Video if event because I would like to get rid of the banner.

What's yours opinion? @mgerhold

PurpurTentakel97:
according to https://github.com/TechStreamConference/website-backend/pull/201
close #210


src/elements/image/video.svelte
```diff
@@ -0,0 +1,32 @@
+<script lang="ts">
+    export let videoPath;
+    export let picturePath;
```
mgerhold, line 3: Shouldn't we use type hints here?

src/routes/dashboard/admin/events/eventsHelper.ts
```diff
@@ -23,28 +24,109 @@ export function convertSaveEventData(data: DashboardEvent): SetAdminEvent {
...
+    // dates
+    if (!isBeforeOrSameDatesString(data.start_date, data.end_date)) {
+        errorList.push('Das Start-Datum liegt nach dem End-Datum.');
+    }
+
+    if (data.publish_date && data.schedule_visible_from) {
+        if (!isBeforeOrSameDatesString(data.publish_date, data.schedule_visible_from)) {
+            errorList.push('Der Ablaufplan ist vor dem Event sichtbar.');
+        }
+    }
```
mgerhold, line 81: Can be simplified.
```suggestion
    if (data.publish_date && data.schedule_visible_from
        && !isBeforeOrSameDatesString(data.publish_date, data.schedule_visible_from)
    ) {
        errorList.push('Der Ablaufplan ist vor dem Event sichtbar.');
    }
```
The same applies to the other `if`s, too.

src/routes/dashboard/admin/events/eventsHelper.ts
```diff
@@ -23,28 +24,109 @@ export function convertSaveEventData(data: DashboardEvent): SetAdminEvent {
...
+        errorList.push('Die Trailer-Poster-URL ist nicht valide.');
+    }
+
+    // speaker
+    for (const speaker of allSpeaker) {
+        if (speaker.visible_from) {
+            if (!isBeforeOrSameDatesString(speaker.visible_from, data.start_date)) {
+                errorList.push(`Speaker ${speaker.name} ist erst nach dem Event-Start sichtbar.`);
+            }
+        }
```
mgerhold, line 124: Same here.

src/forms/adminEventForm.svelte
```diff
@@ -310,14 +203,23 @@
...
-          id="dashboard-admin-event-youtube-trailer-id"
-          labelText="YouTube Trailer ID:"
-          placeholderText="YouTube Trailer ID:"
-          ariaLabel="Gib die ID des YouTube Trailer des ausgewählten Events ein."
-          bind:value={event.trailer_youtube_id}
+          classes="dashboard-admin-event-trailer-url input"
+          id="dashboard-admin-event-trailer-url"
+          labelText="Trailer URL:"
+          placeholderText="Trailer URL:"
+          ariaLabel="Gib die URL des Trailer des ausgewählten Events ein."
```
mgerhold, line 210:
```suggestion
          ariaLabel="Gib die URL des Trailers des ausgewählten Events ein."
```

mgerhold (changes requested):

src/routes/dashboard/admin/events/eventsHelper.ts
PurpurTentakel97: same here.

src/routes/dashboard/admin/events/eventsHelper.ts
mgerhold: I would argue that those nested `if`s are never more readable, because when reading it, the first thought will always be: Why is this nested? Effectively, your only checking a chain of conditions, evaluated in order. The inner `if` (just as a control structure) serves no purpose here. If it's about the readabity, maybe just put every condition in its own line, say:
```ts
    if (
        data.publish_date
        && data.schedule_visible_from
        && !isBeforeOrSameDatesString(data.publish_date, data.schedule_visible_from)
    ) {
        errorList.push('Der Ablaufplan ist vor dem Event sichtbar.');
    }
```
I think this is at least as readable as the original version, but it's not confusion anymore (because nobody will wonder why there are *two* instead of one control structures).

Regarding the other thing you wrote: I don't think you'd need an explicit `null` check. If you're really sure that something is not `null`, you should be able to use the null assertion operator (`!`).

PR #223, opened 2025-06-19
Add `svelte-check` as dependency, update some dependencies, add `npm run check` as pre-commit hook via Husky

mgerhold:


src/pages/year.svelte
```diff
@@ -111,11 +111,16 @@
         </div>
     </div>
     <div class="year-content-wrapper">
-        {#if data.year.event.trailer_url && data.year.event.trailer_poster_url}
+        {#if data.year.event.trailer_url
+        && data.year.event.trailer_poster_url
+        && data.year.event.trailer_subtitles_url}
             <Section id="Trailer">
                 <div class="year-video-wrapper">
                     <Video ariaLabel="Tech Stream Conference Trailer {data.year.event.year}"
```
mgerhold, line 119: We should use the event's name here instead of hard-coding "Tech Stream Conference". The events could have different names.

src/forms/adminTentativeTalkApprovalForm.svelte
```diff
@@ -85,7 +85,7 @@
 
         message.setSaveMessage(result.success ? SaveMessageType.Save : SaveMessageType.Error);
         errorList = result.messages;
-        if (result.success) {
+        if (result.success && talk.suggested_time_slot) { // Time slot cannot be null because the initiating talk needs to have one. To make sure.
```
mgerhold, line 88: I don't understand the comment. You're saying that the time slot cannot be null. And then you say "To make sure". To make *what* sure?

PurpurTentakel97: More to please the type checker. 🙈

PR #253, opened 2025-07-01
Update

PurpurTentakel97:
Please check out if the Frontend does work with this versions.

I needed to extract the enums from the script tags because a error "enum in script tag" occured after updating.
Also I can not update vite to 7.0.0 because this faile while installing Wo we stay on the latest 6.x.x.

I got rid of the warnings.
But I do not unterstand what is wrong with the event forwarding.
AI says something about Event Types are missing. But importing does not help.
Im not quite sure if this is even error or if svelte check just dont get, that `oncropcomplete` is an valid event.

### DIFF
--- a/src/elements/image/cropper.svelte
+++ b/src/elements/image/cropper.svelte
@@ -177,8 +177,11 @@
             return;
         }
 
-        const mouseDeltaX = (event.clientX - lastMouseX) * currentZoomScale;
-        const mouseDeltaY = (event.clientY - lastMouseY) * currentZoomScale;
+        const viewportScaleX = canvas.scrollWidth / canvas.width;
+        const viewportScaleY = canvas.scrollHeight / canvas.height;
+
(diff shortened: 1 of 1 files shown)


mgerhold (approved):
I really like the new cropper now 🥳
I just noticed some minor details:
- There's a blinking cursor within the cropper:
  blinking_cursor
- The speed of the mouse movement still is not perfectly aligned with the movement of the image:
  scroll_movement_offset

But: The cropper is totally usable as of now! :)
I would recommend: Let's merge this PR now. ***If*** you want to address my other suggestions or make other improvements, let's just do it in separate PRs. What do you think?

mgerhold (approved):
I think the cropper (and its code) are now fine 👍

There's one minor thing I noticed, but I don't think you have to fix it. And if you want to fix it, you should maybe do it in a different PR. I noticed that the pivot point of the zoom does not match the mouse position. Or it does match the mouse position, but the current zoom level is not respected. That results in "zooming in a curve". Just see the GIF and note that I don't move the mouse cursor:
scroll_center_offset

PR #275, opened 2025-08-19
270 i cannot use wasd keys in the profile section text boxes

### LINKED ISSUE #270: I cannot use "WASD" keys in the profile section text boxes
I tried changing my name in the profile section, but the W,A,S,D keys were not recognized.

Image

PurpurTentakel97:
close #270 

I added some functions to enable and disable the cropper keyboard events.
By default they are disabled now.


mgerhold (approved):
Works fine on my end 👍

PR #305, opened 2026-01-06
300 restrict lengths of talk titles and descriptions

### LINKED ISSUE #300: Restrict Lengths of Talk Titles and Descriptions
# Summary

We should constrain the maximum length of talk titles and descriptions.

## Titles

When uploading VODs to YouTube, the video title is constrained to a length of 100 characters. We use the following format for the titles:

```
<talk title> – <speaker name> – TECH STREAM CONFERENCE
```

In this format, there are 28 hardcoded characters (` –  – TECH STREAM CONFERENCE`). It would be perfect if we could constrain the length for talk titles to 100 - 28 - `<length of speaker name>` characters. To be safe, there should be a hard-coded minimum length for the title (e.g., 20 characters). On one hand, this feels weird, because the limit would then depend on the speaker’s name. But on the other hand, I think it’s even worse having to shorten the talk title for YouTube.  
If you really don’t like letting the speaker name influence the maximum title length, we should use a reasonable default that works for most speaker names.

Just some statistics: In our current database, the average length of the speaker’s names is between 10 and 11 characters. This would give us a maximum title length of 61. The maximum name length at the moment is 27. This would lead to a maximum title length of 45.

## Descriptions

For talk descriptions, the main constraint is the OBS scene shown between the talks, e.g. this one:

Image

Judging by the data of the last event, I think that a maximum length of 280 or 290 characters would be suitable here (even though it surely depends on the actual text). On the following screenshot, the description in 286 characters long and I think it looks good:

Image

However, we will run into problems with the panel discussions since they usually have descriptions that are way longer. I guess we cannot really do anything about that. But that means: the description length limit should be ignored when admins create new panel discussions (maybe you can add a boolean flag to the properties of that Svelte component to enable/disable the length limit for the description text).

PurpurTentakel97:
close #300

As said in the issue.
We should maybe restrict ourselves while writing a panel discussion description.
The first part of the description could be as long as a speaker description and rather stand alone. 

What do you think?

### DIFF
--- a/src/forms/speakerTalkForm.svelte
+++ b/src/forms/speakerTalkForm.svelte
@@ -8,6 +8,7 @@
     import {trySaveDataAsync} from 'helper/trySaveData';
     import {SaveMessageType} from 'types/saveMessageType';
     import {createEventDispatcher} from 'svelte';
+    import {NO_LIMIT} from 'elements/input/textArea.svelte';
 
     import TagArray from 'elements/input/tagArray.svelte';
     import Message from 'elements/text/message.svelte';
@@ -26,6 +27,7 @@
     export let tags: AllTalkTag;
     export let talkDurations: DashboardTalkDurationChoices;
     export let useForm: boolean = true;
+    export let isAdmin: boolean = false;
 
     // @formatter:off
     let saveMessage: SaveMessage;
@@ -89,7 +91,8 @@
                   ariaLabel="Gib hier die Beschreibung des Talks ein"
                   bind:value={data.description}
                   on:input={setUnsavedChanges}
-                  on:submit={save}/>
+                  on:submit={save}
+                  limit={isAdmin ? NO_LIMIT : 280}/>
         <Paragraph classes="paragraph-gray">Beschreibe deinen Vortrag. Hier darfst du gerne etwas mehr ins Detail gehen.
             Worum geht es in deinem Vortrag? Was wird man dabei lernen? Was ist das
             Besondere an deinem Vortrag? Eine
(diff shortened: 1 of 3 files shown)


mgerhold (changes requested):

src/forms/speakerTalkForm.svelte
```diff
@@ -21,11 +22,16 @@
     import FormWrapper from 'elements/wrapper/formWrapper.svelte';
     import Paragraph from 'elements/text/paragraph.svelte';
 
+    const YOUTUBE_VIDEO_TITLE_LENGTH = 100;
+    const YOUTUBE_VIDEO_TITLE_FIXED_PART = " –  – TECH STREAM CONFERENCE"
```
mgerhold: Hehe 😅 What about your formatter? 🙈

```suggestion
    const YOUTUBE_VIDEO_TITLE_FIXED_PART = " –  – TECH STREAM CONFERENCE";
```

src/forms/speakerTalkForm.svelte
```diff
@@ -21,11 +22,16 @@
     import FormWrapper from 'elements/wrapper/formWrapper.svelte';
     import Paragraph from 'elements/text/paragraph.svelte';
 
+    const YOUTUBE_VIDEO_TITLE_LENGTH = 100;
```
mgerhold: I’m nitpicking here, but actually, it’s not the *length* of the title, but the *maximum length*. So maybe rename it to `YOUTUBE_VIDEO_TITLE_MAX_LENGTH`?  
Also: It’s a bit confusing that we use `LENGTH` for the title, but `LIMIT` for the description. So what about `TALK_DESCRIPTION_MAX_LENGTH` instead of `DESCRIPTION_LIMIT`?

mgerhold (changes requested):

mgerhold (approved):

PR #308, opened 2026-01-06
feat: display duration within the schedule

### LINKED ISSUE #303: Show Talk Durations in Schedule
Inside the schedule, we should not only show the start time of each talk, but also the duration (in minutes), maybe in a second row below the start time.

Note: We decided *against* displaying the end time because it would feel redundant to show both the duration and the end time.

PurpurTentakel97:
close #303

### DIFF
--- a/src/elements/schedule/scheduleEntry.svelte
+++ b/src/elements/schedule/scheduleEntry.svelte
@@ -17,7 +17,7 @@
 
 <div class="schedule-entry-time-wrapper schedule-entry-horizontal-line">
     <SubHeadline classes="schedule-entry-time">{formatDate(talk.starts_at, '%hh:%mm Uhr')}</SubHeadline>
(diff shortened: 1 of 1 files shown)


src/elements/schedule/scheduleEntry.svelte
```diff
@@ -17,6 +17,7 @@
 
 <div class="schedule-entry-time-wrapper schedule-entry-horizontal-line">
     <SubHeadline classes="schedule-entry-time">{formatDate(talk.starts_at, '%hh:%mm Uhr')}</SubHeadline>
+    <TextLine classes="schedule-entry-time">{talk.duration} Min</TextLine>
```
mgerhold:
```suggestion
    <TextLine classes="schedule-entry-time">{talk.duration} Min.</TextLine>
```

